PR #29, opened 2023-11-14
Adds state management and caching to AspireWithDapr Weather sample

paulyuk:
The current sample showcases service invoke + nice treatment of sidecars.

This shows just a bit more around using powerful APIs like state management along with built in integrations with Redis to achieve a better result, simply.

### DIFF
--- a/samples/AspireWithDapr/AspireWithDapr.Web/WeatherApiClient.cs
+++ b/samples/AspireWithDapr/AspireWithDapr.Web/WeatherApiClient.cs
@@ -1,12 +1,29 @@
 ﻿using Dapr.Client;
 
+
 namespace AspireWithDapr.Web;
 
 public class WeatherApiClient(DaprClient daprClient)
 {
     public async Task<WeatherForecast[]> GetWeatherAsync()
     {
-        return await daprClient.InvokeMethodAsync<WeatherForecast[]>(HttpMethod.Get, "api", "weatherforecast");
+    
+       // Get the weather from the state store if it exists
+       var weather = await daprClient.GetStateAsync<WeatherForecast[]>("statestore", "weather");
+
+        if (weather is null)
+        {
+            // If it doesn't exist, get it from the weather service
+            weather = await daprClient.InvokeMethodAsync<WeatherForecast[]>(HttpMethod.Get, "api", "weatherforecast");
+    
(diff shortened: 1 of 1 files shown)


samples/AspireWithDapr/AspireWithDapr.Web/WeatherApiClient.cs
```diff
@@ -1,12 +1,34 @@
...
+    
+       // Get the weather from the state store if it exists
+       var weatherData = await daprClient.GetStateAsync<WeatherForecast[]>(stateStore, "weather");
+
+        if (weatherData is null)
+        {
+            // If it doesn't exist, get it from the weather service
+            weatherData = await daprClient.InvokeMethodAsync<WeatherForecast[]>(HttpMethod.Get, apiAppId, "weatherforecast");
+    
+            await daprClient.SaveStateAsync(stateStore, "weather", weatherData, metadata: new Dictionary<string, string>() {
```
DamianEdwards, line 24: Is the `metadata` parameter typed as `Dictionary<string, string>`? If so, we can simplify to use target-typed `new` here. If not, oh well :smile:

DamianEdwards (approved):

samples/AspireWithDapr/AspireWithDapr.Web/WeatherApiClient.cs
```diff
@@ -1,12 +1,34 @@
...
+    const string stateStore = "statestore";
+    const string stateTTL = "120";
+    const string apiAppId = "api";
+
     public async Task<WeatherForecast[]> GetWeatherAsync()
     {
-        return await daprClient.InvokeMethodAsync<WeatherForecast[]>(HttpMethod.Get, "api", "weatherforecast");
+    
+       // Get the weather from the state store if it exists
+       var weatherData = await daprClient.GetStateAsync<WeatherForecast[]>(stateStore, "weather");
```
eerhardt, line 17: This app also uses output caching at the UI layer. Does it make sense to use both? Should we remove the output caching and use this instead?

DamianEdwards: Oh good point. Technically they're caching at different levels (API response vs. full HTML response) so could still be argued it's valuable.

davidfowl: MOAR Caching

PR #41, opened 2023-11-22
Set dashboards refresh rate

Depechie:
Allow the metric dashboards to auto refresh.
Makes it easier to follow traffic when the auto fetch is activated in the blazor project.


DamianEdwards (approved):

JamesNK (approved):

JamesNK:
Thanks!

PR #42, opened 2023-11-23
[eShopLite] Split EF model into a separate class library

### LINKED ISSUE #39: eshoplite project dotnet publish error:
C:\Program Files\dotnet\sdk\8.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.ConflictResolution.targets(112,5): error NETSDK1152: Found multiple publish output files with the same relative path: D:\cupertino\aspire-samples\samples\eShopLite\eShopLite.CatalogDb\appsettings.De
velopment.json, D:\cupertino\aspire-samples\samples\eShopLite\eShopLite.CatalogService\appsettings.Development.json, D:\cupertino\aspire-samples\samples\eShopLite\eShopLite.CatalogDb\appsettings.json, D:\cupertino\aspire-samples\samples\eShopLite\eShopLite.CatalogService\appsett
ings.json. [D:\cupertino\aspire-samples\samples\eShopLite\eShopLite.CatalogService\eShopLite.CatalogService.csproj]

It seems that both db and service project have appetting file. and output to the same folder/

DamianEdwards:
Splits `CatalogDbContext` and the related model classes into a new class library project that is referenced by both the newly added `eShopLite.CatalogDbManager` app and the existing `eShopLite.CatalogService` app. This removes the "app referencing app" instance which was causing conflicting relative paths on *appsettings.json* during publish of `eShopLite.CatalogService`.

I investigated an MSBuild based solution, suppressing the content files from `eShopLite.CatalogDb` being published with `eShopLite.CatalogService` but the only way I found to do it was via a custom target in `eShopLite.CatalogService` that special-cased removing files from the publish list when they contained 'CatalogDb' and ended with '.json', or by using an undocumented MSBuild property (`_GetChildProjectCopyToPublishDirectoryItems`). You can see that attempt in a2ba5107b47ac904d4b9215c713604ec514c89e3 to understand why I decided not to do it.

Note that this uncovered a slight usability issue with the builder methods provided by the Npgsql EF component, in that you can't directly pass a delegate that configures the `NpgsqlDbContextOptionsBuilder`, so you have to call `UseNpgsql` again. This is needed in this case to set the migrations assembly as [it's separate from the assembly the `DbContext` is in](https://learn.microsoft.com/ef/core/managing-schemas/migrations/projects). We might want to look at adding an overload of `IHostApplicationBuilder.AddNpgsqlDbContext` that accepts an `Action<NpgsqlDbContextOptionsBuilder>` to enable this and other scenarios.

/Cc @BrennanConroy @davidfowl @baronfel @dsplaisted @eerhardt @roji

Fixes #39


samples/eShopLite/eShopLite.CatalogService/CatalogApi.cs
```diff
@@ -52,3 +52,5 @@ public static RouteGroupBuilder MapCatalogApi(this IEndpointRouteBuilder routes)
         return group;
     }
 }
+
+public record Catalog(int FirstId, int NextId, bool IsLastPage, IEnumerable<CatalogItem> Data);
```
eerhardt, line 56: Why is `Catalog` defined in a separate assembly than `CatalogItem`?

samples/eShopLite/eShopLite.CatalogDbManager/Program.cs
```diff
@@ -1,9 +1,12 @@
 ﻿using eShopLite.CatalogDb;
+using eShopLite.CatalogDbManager;
+using Microsoft.EntityFrameworkCore;
 
 var builder = WebApplication.CreateBuilder(args);
 
 builder.AddServiceDefaults();
-builder.AddNpgsqlDbContext<CatalogDbContext>("catalogdb");
+builder.AddNpgsqlDbContext<CatalogDbContext>("catalogdb", null,
+    c => c.UseNpgsql(a => a.MigrationsAssembly(typeof(Program).Assembly.GetName().Name)));
```
eerhardt, line 9: Should we use a better name than `c` and `a` here?

samples/eShopLite/eShopLite.CatalogService/CatalogApi.cs
eerhardt: Maybe it should be called `CatalogItemPage` (or similar) if it is only meant to be a page of data?

samples/eShopLite/eShopLite.CatalogDbManager/Program.cs
DamianEdwards: `optionsBuilder` and `providerOptionsBuilder`?

eerhardt: How about

```suggestion
    builder => builder.UseNpgsql(npgsqlBuilder => npgsqlBuilder.MigrationsAssembly(typeof(Program).Assembly.GetName().Name)));
```
?

DamianEdwards: Don't like the clash of `builder` in the same scope as the app builder. Tricky to come up with good names that aren't just `dbContextOptionsBuilder` and `npgsqlDbContextOptionsBuilder` 😕

eerhardt: ok then maybe:

```suggestion
    optionsBuilder => optionsBuilder.UseNpgsql(npgsqlBuilder => npgsqlBuilder.MigrationsAssembly(typeof(Program).Assembly.GetName().Name)));
```

eerhardt (approved):

PR #52, opened 2023-11-30
Add sample of container with persistent volume mount

### LINKED ISSUE #44: Using a SQL Container with a Volume
We need an example about how to configure SQL Server container with a Volume using Aspire. I have tried using the method `WithVolumeMount`, but with no luck.

DamianEdwards:
Fixes #44


samples/VolumeMount/VolumeMount.AppHost/Program.cs
```diff
@@ -0,0 +1,18 @@
+﻿var builder = DistributedApplication.CreateBuilder(args);
+
+// Using a persistent volume mount requires a stable password for 'sa' rather than the default generated one.
+var sqlpassword = builder.Configuration["sqlpassword"] ??
+    throw new InvalidOperationException("""
+        A password for the local SQL Server container is not configured.
+        Add one to the AppHost project's user secrets with the key 'sqlpassword', e.g. dotnet user-secrets set sqlpassword <password>
+        """);
```
davidfowl, line 8: This is to workaround our bug? Why not generate one?

DamianEdwards: If you generate one each time, then the user created during the first run can't be logged in to during the subsequent runs. At least, that's what I saw.

DamianEdwards: > This is to workaround our bug? 

But yes, in preview.1 you have to work around our bug of the generated passwords not meeting the complexity requirements. I had this originally doing `Guid.NewGuid().ToString()` but found that I'd get database login errors on runs subsequent to the one the migrations are applied to.

samples/VolumeMount/VolumeMount.AppHost/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    throw new InvalidOperationException("""
+        A password for the local SQL Server container is not configured.
+        Add one to the AppHost project's user secrets with the key 'sqlpassword', e.g. dotnet user-secrets set sqlpassword <password>
+        """);
+}
+
+// To have a persistent volume mount across container instances, it must be named (VolumeMountType.Named).
+var database = builder.AddSqlServerContainer("sqlserver", sqlpassword)
+    .WithVolumeMount($"VolumeMount.sqlserver.data", "/var/opt/mssql", VolumeMountType.Named)
```
davidfowl, line 18:
```suggestion
    .WithVolumeMount("VolumeMount.sqlserver.data", "/var/opt/mssql", VolumeMountType.Named)
```

samples/VolumeMount/VolumeMount.AppHost/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    throw new InvalidOperationException("""
+        A password for the local SQL Server container is not configured.
+        Add one to the AppHost project's user secrets with the key 'sqlpassword', e.g. dotnet user-secrets set sqlpassword <password>
+        """);
+}
+
+// To have a persistent volume mount across container instances, it must be named (VolumeMountType.Named).
+var database = builder.AddSqlServerContainer("sqlserver", sqlpassword)
+    .WithVolumeMount("VolumeMount.sqlserver.data", "/var/opt/mssql", VolumeMountType.Named)
```
mitchdenny, line 18: This differs a little bit from the guidance we see in the SQL docs:

https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-docker-container-configure?view=sql-server-ver16&pivots=cs1-bash

That said it probably works because we aren't blowing anything away. When you run this do you see anything other than stuff in the data, secrets, and logs directories?

mitchdenny:
I like this. It's a good approach to solving the persistent container issue.

samples/VolumeMount/VolumeMount.AppHost/Program.cs
DamianEdwards: Hmmm it seems to match from my reading? Am I not reading carefully enough? Passing a string that's not a path results in Docker creating a named volume which it manages. Am I missing something?

mitchdenny: I beg your pardon it is me who isn't reading clearly :)

samples/VolumeMount/VolumeMount.AppHost/Program.cs
mitchdenny: You know to get around the persistent password problem (generically) we could look at capturing and storing the generated password so that it is reused on subsequent executions of the AppHost. That way folks wouldn't need to worry about doing these config shenanigans (unless they really want to).

DamianEdwards: I like the idea. I can see one challenge being that the logical place to store it would be user secrets but that requires mutating the AppHost project (it requires a property to be set to set the user secrets ID), and mutating the project file automatically while the app is running is probably not a great idea.

PR #77, opened 2024-01-04
Add OrchardCore CMS sample

MikeAlhayek:
When this Aspire sample run, you'll have access to OrchardCore CMS app in addition to an OrchardCore MVC app.


MikeAlhayek:
@sebastienros

sebastienros:
I think it would have more value to register Redis, a database (postgres?) and optionally another one with cloud services (azure storage, ai search) that Orchard supports to show how Aspire helps provisioning all these automatically. Starting two unrelated apps (one being a sample mvc project) doesn't add much value of the combination of OC and Aspire.

It's nice to see that it's already working seamlessly.

MikeAlhayek:
@sebastienros adding `(azure storage, ai search)` requires an account to access these services. I can add database services so when the CMS run, the user can just select a database and setup the site

sebastienros:
I know, I hinted about two projects, in case someone would still want to try it by providing custom keys, or try the deployment.

> the user can just select a database and setup the site

Ideally the connection string that the host generates would be used in the OC app automatically.

PR #108, opened 2024-02-07
Update samples for preview.3

### LINKED ISSUE #90: [WebToolsE2E][Aspire] After updating to Aspire 8.0 Preview 3, opening eShopLite.sln fails with: "Package downgrade detected: Microsoft.Extensions.Options from 8.0.1 to centrally defined 8.0.0"
**REGRESSION INFO**: Worked fine with Aspire 8.0.0-preview.2.23619.3

**INSTALL STEPS**

1. Clean machine: Win11 x64 23h2 ENU
2. Install VS 17.10 Preview 1[34511.213.main]
3. Install Aspire Preview 3 latest build 8.0.0-preview.3.24064.1
4. Apply NuGet Feeds
5. Install Docker Desktop

**REPRO STEPS**

1. Clone aspire-samples repo from main branch
2. Update NuGet.config to use the dotnet8 feed
 image
3. Update Directory.Packages.props to use aspire version 8.0.0-preview.3.24064.1
4. Update global.json to use sdk version 8.0.200-preview.24060.48
6. Open solution eShopLite.sln

**NOTE**

- This issue also reproduces on Metrics / AspireWithNode solutions.

**ACTUAL**

NU1109	Detected package downgrade: Microsoft.Extensions.Options from 8.0.1 to centrally defined 8.0.0. Update the centrally managed package version to a higher version. 
 eShopLite.ServiceDefaults -> Microsoft.Extensions.ServiceDiscovery.Abstractions 8.0.0-preview.3.24064.1 -> Microsoft.Extensions.Options (>= 8.0.1) 
 eShopLite.ServiceDefaults -> Microsoft.Extensions.Options (>= 8.0.0)
image

**EXPECTED**

Opening solution without any errors.

**OTHER INFO**
image

balachir:
WIP branch for updating samples to preview.3. Make changes and/or additions here as necessary.

Fixes #90

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.100",
+    "version": "8.0.200",
     "rollForward": "latestPatch",
     "allowPrerelease": true
   }
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.200-preview.23624.5",
+    "version": "8.0.200",
     "rollForward": "latestPatch",
     "allowPrerelease": true
   }
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.200",
+    "version": "8.0.200-preview.23624.5",
     "rollForward": "latestPatch",
     "allowPrerelease": true
   }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
     - name: Dapr tool installer
       uses: dapr/setup-dapr@v1
       with:
-        version: '1.12.5'
+        version: '1.12.0'
 
     - name: Install .NET SDK
       uses: actions/setup-dotnet@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,3 +43,10 @@ jobs:
 
     - name: Build
       run: ${{ matrix.os == 'ubuntu-latest' && './build.sh' || '.\build.cmd' }}
+
+    - name: Archive Build Logs
+      if: always()
+      uses: actions/upload-artifact@v1
+      with:
+        name: build-results
+        path: '**/*.msbuild'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,4 +50,4 @@ jobs:
       with:
         name: build-log
         path: |
-          **/*.binlog
+          *.binlog
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,4 +50,4 @@ jobs:
       with:
         name: build-log
         path: |
-          **/*.msbuild
+          **/*.binlog
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
     name: Build Samples
     env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,8 @@ jobs:
 
     - name: Dapr tool installer
       uses: dapr/setup-dapr@v1
+      with:
+        version: '1.12.5'
 
     - name: Install .NET SDK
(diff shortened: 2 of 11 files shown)


DamianEdwards:
This won't pass the checks until preview.3 is released in its current state as it's not using the release feeds or a preview 8.0.200 SDK release. @balachir do you want to try updating the **nuget.config** to use the release feed and the **global.json** to point at the latest 8.0.200-preview SDK (8.0.200-preview.23624.5) for now to ensure this builds?

DamianEdwards:
> This won't pass the checks until preview.3 is released in its current state as it's not using the release feeds or a preview 8.0.200 SDK release. @balachir do you want to try updating the nuget.config to use the release feed and the global.json to point at the latest 8.0.200-preview SDK (8.0.200-preview.23624.5) for now to ensure this builds?

OK I ended up doing the above myself and now we can see what actually needs to be updated in this PR due to changes in preview.3

ReubenBond (approved):

ReubenBond:
I believe this just needs the NuGet.config change reverted before merging

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.100",
+    "version": "8.0.200-preview.23624.5",
```
ReubenBond, line 3: Does this need to be updated?

balachir:
> I believe this just needs the NuGet.config change reverted before merging
I've reverted the NuGet.config changes now.

global.json
balachir: Yes, I was wondering about it too. Should we set the version to 8.0.200 now that the SDK is available? I'm not sure if the build agent is using this 8.0.200-preview version.

ReubenBond: I think so, yes.

ReubenBond: I pushed a commit to update it

ReubenBond:
Do we need to add `dotnet workload update && dotnet workload install aspire` to the workflow yaml?

DamianEdwards:
> Do we need to add dotnet workload update && dotnet workload install aspire to the workflow yaml?

It's already in the `build.sh`

DamianEdwards:
The failure repros locally, seems there's an issue generating the manifest for the Dapr sample.

ReubenBond:
Ah, the issue is that the Dapr CLI is missing on the CI machine:

ReubenBond:
I added a Dapr CLI installation step to the pipeline

DamianEdwards:
@ReubenBond still failing

ReubenBond:
Yeah, I wonder if the dapr integration isn't working on Windows (which is running first and cancelling the linux build) - I'll see if it works on Linux and we can hold off on merging for now

ReubenBond:
Ok, that wasn't it. We get a different error on Linux. I tried updating to the latest DAPR CLI (the default from the action is very old). I assume it's still not finding the binary. I am trying to enable MS Build Structured Logs to the CI to see what the exact issue is now. We might end up needing to set the dapr/path env var manually

ReubenBond:
Looking at our code now... did this ever work? The dapr hosting package in Aspire *only* looks in `C:\dapr\dapr.exe` on Windows. For Linux, it checks `~/dapr/dapr` and `/usr/local/bin/dapr`. The `setup-dapr` task doesn't support selecting the installation directory (it outputs a `dapr-path` var).

I think we need to change the Aspire.Hosting.Dapr package to either check the PATH or allow specifying the Dapr installation directory via an env var

timheuer:
I think ServiceDefaults package needs a change to make the package ref to Http.Options instead of Resliency.

ReubenBond:
I'm tempted to say that we should either merge without CI or disable the Dapr sample from building in CI until https://github.com/dotnet/aspire/issues/2219 is resolved and we have updated to a version with that (i.e, Aspire preview 4).
cc @philliphoff - perhaps you know how we might resolve this

davidfowl:
Skip the tests for the dapr sample

joperezr:
Looks like this is not updating all of the packages that were shipped yesterday, instead it is only updating the aspire ones. I can push another commit to make sure everything is up-to-date

joperezr (approved):
Looks like this is ready to go.

philliphoff:
> perhaps you know how we might resolve this

@ReubenBond An ugly way would be to set the `DaprOptions.DaprPath` value to the CLI path found in `PATH` in the application host when adding Dapr support in the sample.  I've never looked at the GH Action before; I didn't realize it dumped it in a random directory.  Agree that it could be useful to use what's found in PATH, if anything, before looking at other specific locations.

PR #119, opened 2024-02-20
Add project file for JavaScript projects in AspireJavaScript solution

PranavSenthilnathan:
Adds an esproj file to AspireJavaScript.sln so that the JavaScript projects can be viewed in Visual Studio:
image


timheuer:
>this PR instead directly sets the backend and frontends as startup projects

The one problem with this is the AppHost is critical to the orchestration process and sets the environment variables that are needed between different app types. So I think we'd actually want to ensure that the AppHost/DCP can launch the esproj types...

DamianEdwards:
@timheuer seems we should close this PR for now until such time that we actually support esproj projects properly?

PR #121, opened 2024-02-21
Fix entrypoint for Vue app container

timheuer:
When deploying the javascript sample with the Vue app, the container failed to start with:

```
Unknown command: "dev"
Did you mean this?
npm run dev # run the "dev" package script
```

it appears that in the container the implied script name (unlike start) is not known.


IEvangelist:
This is invalid after #125 is approved/merged.

PR #124, opened 2024-02-21
macosx gitignore

moljac:
Just extra patterns to be ignored on MacOSX

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@
 ##
 ## Get latest from https://github.com/github/gitignore/blob/main/VisualStudio.gitignore
 
+# MacOSX files
+**/.DS_Store
+**/._*
+
 # User-specific files
 *.rsuser
 *.suo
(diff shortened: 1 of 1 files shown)


DamianEdwards:
Thanks for this. Thinking about who can review this. Tagging some macOS users I know @ellismg @captainsafia

moljac:
@DamianEdwards 

Take your time. I work mostly on Macs, so we can cover as much platforms as possible and those files are a bit annoying.

I am investigating clients (MAUI as no.1) and will ping you soon with tons of questions.

captainsafia:
This is a good minimum set for macOS. If you want to be extra thorough, I'd recommend copying the [macOS.gitignore template from GitHub's catalog](https://github.com/github/gitignore/blob/main/Global/macOS.gitignore).

DamianEdwards:
Thanks @captainsafia! @moljac would you like to copy those over too?

moljac:
> Thanks @captainsafia! 

Thanks. 

> @moljac would you like to copy those over too?

sure.

DamianEdwards (approved):

PR #125, opened 2024-02-21
Fix JavaScript container images

### LINKED ISSUE #122: Angular Docker image doesn't respond when run as container
Unclear if this is an azd or Angular sample issue, but when deploying the AspireWithJavaScript sample to Azure using azd, all the front-ends work except the Angular one, it fails to respond. No errors in log stream and appears to start, so I'm not sure where the issue lies yet, but logging the issue here to start.

IEvangelist:
Fix `docker run` issue with Angular container image. I needed the `--host 0.0.0.0` bits. While doing this, I realized that the container images were massive and fixed that by re-layering atop `nginx:alpine`:

| JavaScript container | Original size | New size |
|---|---|---|
| **Angular** | 1.68 GB | 42 MB |
| **React** | 1.69 GB | 43 MB |
| **Vue** | 1.32 GB | 42 MB |

- Fixes #122
- Closes #121


timheuer (changes requested):
Need npm install commands before build

IEvangelist:
> Need npm install commands before build

All of the _Dockerfile's_ do have a `RUN npm i`, see line 8 of each file.

timheuer:
My bad, diff view was hiding it!

The Angular image when deployed is 404 on the API call for some reason still... :-(

image

Like the proxy replacement isn't working when deployed in ACA

timheuer:
Got it!  Add this:

in AspireJavaScript.Angular/Dockerfile: add
`COPY --from=build /app/nginx-default.conf.template /etc/nginx/templates/default.conf.template`
after the app copy

Then add a file named `nginx-default.conf.template` into the AspireJavaScript.Angular folder with this contents:
```
server {
    listen       80;
    listen  [::]:80;
    server_name  localhost;

    #access_log  /var/log/nginx/host.access.log  main;

    location / {
        root   /usr/share/nginx/html;
        index  index.html index.htm;
    }

    #error_page  404              /404.html;

    # redirect server error pages to the static page /50x.html
    #
    error_page   500 502 503 504  /50x.html;
    location = /50x.html {
        root   /usr/share/nginx/html;
    }

    location /api {
        proxy_pass ${services__weatherapi__1};
        proxy_http_version 1.1;
        proxy_ssl_server_name on;
        rewrite ^/api(/.*)$ $1 break;
    }
}


```

IEvangelist:
> Got it! Add this:
> 
> in AspireJavaScript.Angular/Dockerfile: add `COPY --from=build /app/nginx-default.conf.template /etc/nginx/templates/default.conf.template` after the app copy
> 
> Then add a file named `nginx-default.conf.template` into the AspireJavaScript.Angular folder with this contents:
> 
> ```
> server {
>     listen       80;
>     listen  [::]:80;
>     server_name  localhost;
> 
>     #access_log  /var/log/nginx/host.access.log  main;
> 
>     location / {
>         root   /usr/share/nginx/html;
>         index  index.html index.htm;
>     }
> 
>     #error_page  404              /404.html;
> 
>     # redirect server error pages to the static page /50x.html
>     #
>     error_page   500 502 503 504  /50x.html;
>     location = /50x.html {
>         root   /usr/share/nginx/html;
>     }
> 
>     location /api {
>         proxy_pass ${services__weatherapi__1};
>         proxy_http_version 1.1;
>         proxy_ssl_server_name on;
>         rewrite ^/api(/.*)$ $1 break;
>     }
> }
> ```
> 
> with these changes the container is running fine locally -- deploying now to see... UPDATE: 😢 fail when running on ACA still

That's really close to what I had, will try that when I'm back to my machine. I'm curious, what's the error now that we've gotten this far once deployed?

timheuer:
>what's the error now that we've gotten this far once deployed?

This is the frustrating thing.  When using the *same container image* locally and the URI for the environment it works.  It's just in the ACA environment it doesn't.  ANd there is nothing on the server logs that I can see.  Just 404 from the client still.

IEvangelist:
> > what's the error now that we've gotten this far once deployed?
> 
> This is the frustrating thing. When using the _same container image_ locally and the URI for the environment it works. It's just in the ACA environment it doesn't. ANd there is nothing on the server logs that I can see. Just 404 from the client still.

Maybe we're approaching this the wrong way. I mean, really, why do these need to be containerized? They're simple static web apps, would it be better to deploy these as an [Azure Static Web App](https://learn.microsoft.com/azure/static-web-apps/overview)?

timheuer:
Maybe, but I'm trying to validate the e2e for these types of apps within the Aspire workflow...which involves at a minimum knowing the service endpoints...something that doesn't get set right now until deployment.  Lots of manual ways to do it, but again trying to validate that what you put in the AppHost is deployable to our ACA environment.

aaronpowell:
I've done some digging in here and there's a few steps that are missing to get Angular working.

First off, we'll need to include the nginx.config template that @timheuer has above, since we need to use `proxy_pass` to get the API requests forwarded through to the service backend.

To use the environment variable from aspire, we need to do a subsitute with `envsubst` as part of the Dockerfile `CMD` directive, updating it like this should work:

`envsubst < /etc/nginx/templates/nginx-default.conf.template > /etc/nginx/conf.d/default.conf && exec nginx -g 'daemon off;'`

This has worked on my local testing

IEvangelist:
> I've done some digging in here and there's a few steps that are missing to get Angular working.
> 
> First off, we'll need to include the nginx.config template that @timheuer has above, since we need to use `proxy_pass` to get the API requests forwarded through to the service backend.
> 
> To use the environment variable from aspire, we need to do a subsitute with `envsubst` as part of the Dockerfile `CMD` directive, updating it like this should work:
> 
> `envsubst < /etc/nginx/templates/nginx-default.conf.template > /etc/nginx/conf.d/default.conf && exec nginx -g 'daemon off;'`
> 
> This has worked on my local testing

Yes, I had all that and a few more changes I've been working on locally. But, it doesn't work. I'll push a commit with the changes that I have locally, and we can go from there to see if I'm missing something.

timheuer:
Just FYI the envsubst is automatic now in nginx, so you just have to copy to the templates dir and it moves it to conf.d for you.  Agreed though let's see what the ACA team can help us with responding to the 'why doesn't it work there' answer

timheuer (commented):
Won't this fail locally though? you are trying to bind every app to same port?

samples/AspireWithJavaScript/AspireJavaScript.Angular/default.conf.template
```diff
@@ -0,0 +1,17 @@
+server {
+    listen       80;
```
timheuer, line 2: I would change these to 3000

samples/AspireWithJavaScript/AspireJavaScript.Angular/Dockerfile
```diff
@@ -9,6 +9,13 @@ RUN npm i
...
+RUN npm run build-prod
 
-CMD [ "npm", "start" ]
\ No newline at end of file
+FROM nginx:alpine
+
+COPY --from=build /app/default.conf.template /etc/nginx/templates/default.conf.template
+COPY --from=build /app/dist/weather/browser /usr/share/nginx/html
+
+EXPOSE 80
```
timheuer, line 19: Change to 3000

samples/AspireWithJavaScript/AspireJavaScript.AppHost/Program.cs
```diff
@@ -6,19 +6,19 @@
 // Angular: npm run start
 builder.AddNpmApp("angular", "../AspireJavaScript.Angular")
     .WithReference(weatherApi)
-    .WithEndpoint(containerPort: 3000, scheme: "http", env: "PORT")
+    .WithHttpEndpoint(containerPort: 80, env: "PORT")
```
timheuer, line 9: I would change these back to 3000,3001,3002

samples/AspireWithJavaScript/AspireJavaScript.React/Dockerfile
```diff
@@ -9,6 +9,12 @@ RUN npm i
...
-EXPOSE 3000
+RUN npm run build
 
-CMD [ "npm", "start" ]
\ No newline at end of file
+FROM nginx:alpine
+
+COPY --from=build /app/build /usr/share/nginx/html
+
+EXPOSE 80
```
timheuer, line 18: Change to 3001

samples/AspireWithJavaScript/AspireJavaScript.Vue/Dockerfile
```diff
@@ -9,6 +9,12 @@ RUN npm i
...
-EXPOSE 5173
+RUN npm run build
 
-CMD [ "npm", "dev" ]
\ No newline at end of file
+FROM nginx:alpine
+
+COPY --from=build /app/dist /usr/share/nginx/html
+
+EXPOSE 80
```
timheuer, line 18: 3002

timheuer (changes requested):
I think changing back to 3000/1/2 would be better and then just change the docker exposes here.  Did this and all works for me now.

IEvangelist:
We'll have to update this PR when [this issue](https://github.com/dotnet/aspire/issues/2441) is addressed with a means of providing `--build-arg` values.

DamianEdwards:
@IEvangelist the issue you linked to above is on the backlog now and out of scope for the meantime. Is there anything still needed to be done to fix/improve the AspireWithJavaScript sample or should we close this PR for now?

IEvangelist:
> @IEvangelist the issue you linked to above is on the backlog now and out of scope for the meantime. Is there anything still needed to be done to fix/improve the AspireWithJavaScript sample or should we close this PR for now?

@DamianEdwards - This PR will be needed when that linked issue gets fixed, I'm fine closing this for now and revisiting later. But I'm also fine with just sitting and coming back to it once the underlying issue is addressed. Up to you, which do you prefer?

DamianEdwards:
@IEvangelist I don't really like leaving PRs open for indeterminate amounts of time. There's no knowing when that issue will be addressed, if ever, so I prefer to close the PR until such time that there's active work to do.

PR #150, opened 2024-03-10
Fixed subscribe call in angular sample

sgbj:
Fixed a syntax issue where the previous code tried to pass an observer to the [`subscribe`](https://rxjs.dev/api/index/class/Observable#subscribe) method but passed an anonymous function instead. This caused the curly braces to become the body of the function and `next` and `error` to become labels.


DamianEdwards:
FYI @IEvangelist

IEvangelist (approved):

PR #198, opened 2024-04-16
Enable dependabot

### LINKED ISSUE #197: enable dependabot
Eg., this would automatically push the OTEL libraries to 1.8.

I'm trying this out in my fork first.

danmoseley:
Fixes #197

The test run in my fork timed out, but it successfully created an update PR, and such timeouts are sporadic right now in other repos eg dotnet/aspire and generally invisible to those working in the repo.

Note, without the uber-sln, Dependabot fails immediately with  "Dependabot couldn't find a *.(sln|csproj|vbproj|fsproj|proj)". I do not know how to resolve this in another way. The uber-sln is not useful to open in VS. cc @brettfo


DamianEdwards:
Is there a way we can hide the uber solution in a sub-directory (e.g. `build`) and point dependabot at it via config, so it's not in the repo root?

danmoseley:
@brettfo any guidance here? I find the dependabot code hard to follow for the unfamiliar.

@DamianEdwards Might be worth merging as is if we don't have an immediate answer, just to get packages updated near term.

.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+registries:
+  public-nuget:
+    type: nuget-feed
+    url: https://api.nuget.org/v3/index.json
+updates:
+  - package-ecosystem: nuget
```
DamianEdwards, line 7: What about npm? We have Node.js samples too.

danmoseley: good point -- let me experiment

danmoseley: just for the direct package.json presumably, not the ones in node_modules as well.

danmoseley:
>> Is there a way we can hide the uber solution in a sub-directory (e.g. build) and point dependabot at it via config, so it's not in the repo root?
>
> @brettfo any guidance here? I find the dependabot code hard to follow for the unfamiliar.

For the record, if I move the sln out of the root, I get 
> Dependabot couldn't find a *.(sln|csproj|vbproj|fsproj|proj)
Dependabot requires a *.(sln|csproj|vbproj|fsproj|proj) to evaluate your .NET dependencies. It had expected to find one at the path: /*.(sln|csproj|vbproj|fsproj|proj).

even if I put it in say `/dependabot` and have
```
  - package-ecosystem: nuget
    directory: "/dependabot"
```

I can't follow what dependabot is doing easily as this comes from the ruby part
https://github.com/dependabot/dependabot-core/blob/2e9903a4a5bf057def43f2890ba3d0fe3ee02185/nuget/lib/dependabot/nuget/file_fetcher.rb#L12
but then subsequently projects are discovered in the C#
https://github.com/dependabot/dependabot-core/blob/2e9903a4a5bf057def43f2890ba3d0fe3ee02185/nuget/helpers/lib/NuGetUpdater/NuGetUpdater.Core/Discover/DiscoveryWorker.cs#L128
I'm wondering whether the "directories" part is being ignored by the ruby part.
Anyway, archiving this here for now...

brettfo:
The `directory` property _should_ be honored (at least it was in the past), so if it's not that's certainly a bug.  I'm looking at some timeout issues currently then I'll look at the `directory` property again.

Issue #192, opened 2024-04-15
NuGet package versioning is hidden from sample users


yoDon:
Never, ever, not in a million years, would I have thought to look for an inherited package manager versioning specifier document somewhere above the otherwise complete sample solution folder I care about. 

Please don't do "clever" things like this in sample repos.

Users, like me, are going to attempt to build their first Aspire project by copying one of the sample solutions out of this repo and pasting it into their own repo to start messing with it.

The copied sample won't work.

None of the samples will work.

Hours will be wasted, and the devs will either give up or come close to it.

davidfowl:
cc @DamianEdwards

DamianEdwards:
Yep agreed. We discussed this when first creating this repo but the decision was made to use central package management (CPM). Now that we're getting Dependabot configured I'll remove CPM from this repo for the reasons you cite.

PR #199, opened 2024-04-17
update dependabot #2

danmoseley:
Replacing https://github.com/dotnet/aspire-samples/pull/198 since I deleted my fork to test dependabot clean. (Maybe I should make a 2nd account just for this purpose)

Now updating NPM (and github actions) as well as Nuget. NPM ones are aggregated per-sample because merges of the lock file are hopeless.

On the test run, it timed out on nuget, but made PR's for the other two kinds: https://github.com/danmoseley/aspire-samples/pulls  .. my experience is that it will likely successfully make nuget PR's on a future run.

I haven't found a way to successfully move the .sln, but maybe that will emerge in future.


DamianEdwards (approved):
Thanks. I'm a bit concerned that the NPM samples have to be listed explicitly as that will easily become out of sync if any are added or removed in the future but I don't think we should block on that for now.

danmoseley:
Right, I would need a dependabot expert to tell me whether there's a way to avoid it. It seems like it does require each package.lock explicitly.

danmoseley:
BTW, the nuget leg is consistently timing out. A bigger problem than this repo. But it means that we should assume we own updating nuget dependencies. I'll have a shot at doing this manually now.

DamianEdwards:
> BTW, the nuget leg is consistently timing out. A bigger problem than this repo. But it means that we should assume we own updating nuget dependencies. I'll have a shot at doing this manually now.

We assume that for now you mean? Otherwise, why did we add dependabot for NuGet at all?

Also we should fix #192 while we're doing this and remove CPM altogether.

danmoseley:
>  Otherwise, why did we add dependabot for NuGet at all?

Because it periodically comes to life. We have been wrestling with this issue, with help from @brettfo , on dotnet/aspire and dotnet/eshop as well. It has gotten worse, but it's not caused by us; when it's fixed, dependencies will start updating without another PR. Meantime, the timing out is invisible to us.

PR #210, opened 2024-04-17
update nuget packages

danmoseley:
Pick up the otel CVE fix, plus update others since dependabot is struggling right now.

Used `dotnet outdated -vl Major -u for.dependabot.only.sln` (installed from https://github.com/dotnet-outdated/dotnet-outdated?tab=readme-ov-file#installation)


danmoseley:
I can fix https://github.com/dotnet/aspire-samples/issues/192 after this, meantime I suggest merge this so the version change isn't mixed in.

DamianEdwards (approved):

PR #211, opened 2024-04-17
Add sample AspireTurboMonoRepo

yoDon:
This sample demonstrates a .NET Aspire application hosting several Node apps organized in a turbo monorepo format. 

The sample is based on [the official Vercel Turbo Kitchen Sink](https://vercel.com/templates/remix/turborepo-kitchensink)


yoDon:
@dotnet-policy-service agre

yoDon:
@dotnet-policy-service agree

yoDon:
@dotnet-policy-service agree

DamianEdwards:
Thanks for the PR. However, I don't think this sample is a good fit for this repo based on our [contributing guidelines](https://github.com/dotnet/aspire-samples?tab=readme-ov-file#contributing), given it doesn't appear to add anything additional to the existing samples for Node.js and JavaScript frameworks with regards to demonstrating actual features of Aspire. I'd suggest this sample better belongs in its own repo or an existing repo associated with the Vercel Turborepo Kitchen Sink Starter.

Happy to discuss further if you feel I'm missing something.

yoDon:
@DamianEdwards A couple thoughts on why this sample is useful... I believe:

1) It's the only sample that shows how to use `yarn` or `pnpm`, which is important because it's not trivial to do that at present with Aspire.

2) it's the only sample that shows how to use an explicit build step for JavaScript/TypeScript/Node projects, including the only one that shows how to use a build step as required for a JS/TS mono-repo like a turbo repo (this is also not trivial to do, since there is no support for explicit builds in the Aspire app model today).

3) once there is future support for explicit builds in the Aspire app model, it becomes a natural sample to update to show how to do that (when/if Aspire adds that capability).

DamianEdwards:
@yoDon could we perhaps instead update one of the existing apps in the JavaScript example to show using `yarn`/`pnpm` and integrating an explicit build step? Maintaining another JavaScript sample (especially one tied to another sample that is sure to continue evolving on its own) is something I'd like to avoid unless it's showing something really different.

yoDon:
I understand - works for me

PR #214, opened 2024-04-19
Disable parallel publish

sebastienros:
Attempting to mitigate https://github.com/dotnet/runtime/issues/101154

### DIFF
--- a/build/Build.proj
+++ b/build/Build.proj
@@ -26,7 +26,7 @@
         <Message Text="Publishing @(SampleSln->Count()) samples" Importance="High" />
 
         <MSBuild Projects="@(SampleSln)" Targets="Restore" />
-        <MSBuild Projects="@(SampleSln)" Targets="Publish" BuildInParallel="True" />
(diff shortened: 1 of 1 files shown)


DamianEdwards:
Still seems to be failing with the file in use error 😞

DamianEdwards:
@sebastienros how about we just disable Web CIL in `MetricsApp.Client` project for now via `WasmEnableWebcil=false`

maraf:
The `Target Name=GeneratePublishBlazorBootExtensionJson Project=MetricsApp.Client.csproj` is being run twice
- As part of MetricsApp.Client.csproj publish
- As part of MetricsApp.csproj publish when computing static web assets for referenced projects.

All solution projects are published in parallel
image

[msbuild.binlog.zip](https://github.com/dotnet/aspire-samples/files/15044451/msbuild.binlog.zip)

Publishing only the server projects seems like a solution

@javiercn Have you ever tackled this?

PR #226, opened 2024-04-24
Update the AspireWithJavaScript bits

### LINKED ISSUE #182: Update AspireWithJavaScript sample to not use create-react-app
create-react-app is deprecated and not being updated and already is in a state that results in warnings and vulnerabilities. We should update the AspireWithJavaScript so that the react app is not using create-react-app anymore.

/Cc @IEvangelist

### LINKED ISSUE #216: Angular app in AspireWithJavaScript sample does not work after publishing to Azure Container Apps
After publishing to Azure Container Apps with azd, the Angular container fails to start with the following error:

```
024-04-19T17:18:45.604499792Z 
2024-04-19T17:18:45.604530960Z > angular-weather@0.0.0 start
2024-04-19T17:18:45.604535479Z > run-script-os
2024-04-19T17:18:45.604538124Z 
2024-04-19T17:18:46.407676426Z 
2024-04-19T17:18:46.407691104Z > angular-weather@0.0.0 start:default
2024-04-19T17:18:46.407694300Z > ng serve --port $PORT
2024-04-19T17:18:46.407696905Z 
2024-04-19T17:18:48.402502480Z - Building...
2024-04-19T17:18:48.402522388Z 
2024-04-19T17:18:48.413545050Z ✘ [ERROR] Cannot start service: Host version "0.19.5" does not match binary version "0.20.1"
2024-04-19T17:18:48.413557172Z 
2024-04-19T17:18:48.413561009Z 1 error
2024-04-19T17:18:48.426148012Z An unhandled exception occurred: The service was stopped: write EPIPE
2024-04-19T17:18:48.426160355Z See "/tmp/ng-eUHaTg/angular-errors.log" for further details.
```

IEvangelist:
In this PR:

- I rewrote the React app from the ground up, removing the `create-react-app` bits.
- Added top-level _.gitignore_ rule for `azd` files.
- Bump all npm package versions.
- Bump _Dockerfile_ `FROM node` versions, now on 20.12.
- Update _/images/_ where applicable.
- Reimagined how the angular app operates, instead serve with `nginx`. Only applied this flow to angular for now, as it was simple, whereas the other SPAs would require more work.

Functions locally as it always has, but now works beautifully with `azd` — here are the three deployments (will delete once this is merged):

- [Vue Weather](https://vue.yellowtree-fed0ffc7.centralus.azurecontainerapps.io/)
- [React Weather](https://react.yellowtree-fed0ffc7.centralus.azurecontainerapps.io/)
- [Angular Weather](https://angular.yellowtree-fed0ffc7.centralus.azurecontainerapps.io/)

Issues addressed in this PR:

- Fixes #182
- Fixes #216


samples/AspireWithJavaScript/AspireJavaScript.Angular/Dockerfile
```diff
@@ -9,4 +9,14 @@ RUN npm install
...
+
+FROM nginx:alpine
+
+COPY --from=build /app/default.conf.template /etc/nginx/templates/default.conf.template
+COPY --from=build /app/dist/weather/browser /usr/share/nginx/html
+
+# Expose the default nginx port
+EXPOSE 80
+
+CMD ["nginx", "-g", "daemon off;"]
```
DamianEdwards, line 22: This is nice! Would be much better if there were an easy way to deploy a built SPA app (aka a bundle of static files) onto a YARP-based container image 🤔

samples/AspireWithJavaScript/AspireJavaScript.React/webpack.config.js
```diff
@@ -0,0 +1,54 @@
...
+      }),
+      new webpack.DefinePlugin({
+        "process.env": {
+          "REACT_APP_WEATHER_API_HTTPS": JSON.stringify(
+            process.env.REACT_APP_WEATHER_API_HTTPS
+          ),
+          "REACT_APP_WEATHER_API_HTTP": JSON.stringify(
+            process.env.REACT_APP_WEATHER_API_HTTP
+          ),
+        }
```
DamianEdwards, line 28: Why is it necessary to redefine these? The hope was setting them from the AppHost directly would negate the need to do this kind of mapping.

IEvangelist: Without this remapping as a webpack plugin, the `process.env.REACT_APP_WEATHER_API_HTTP` isn't actually available from within the app like we're using it. I could have mapped it to a new const value of `MY_API` and within the app used that instead. For example, if I were to uncomment this config and run the app - it errors out with:

```
Uncaught ReferenceError: process is not defined
    at eval (index.js:15:17)
    at ./src/index.js (bundle.js:41:1)
    at __webpack_require__ (bundle.js:569:32)
    at bundle.js:1656:37
    at bundle.js:1658:12
```

We are essentially faking that `process.env` is available with this plugin, perhaps it's clearer to map it to a non-`process.env` name. I'll give that a go.

samples/AspireWithJavaScript/AspireJavaScript.Angular/package.json
```diff
@@ -7,8 +7,8 @@
   "scripts": {
     "ng": "ng",
     "start": "run-script-os",
-    "start:win32": "ng serve --port %PORT%",
-    "start:default": "ng serve --port $PORT",
+    "start:win32": "ng serve --port %PORT% --host 0.0.0.0",
+    "start:default": "ng serve --port $PORT --host 0.0.0.0",
```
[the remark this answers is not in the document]
IEvangelist: Yeah, that's not needed anymore since I went the `nginx` path. I'll remove it.

samples/AspireWithJavaScript/AspireJavaScript.React/webpack.config.js
DamianEdwards: Right, but I thought that environment variables that started with `REACT_APP_` were auto-propagated to a structure that was available from the app. I thought that's how I had it working with the last change.

samples/AspireWithJavaScript/AspireJavaScript.Angular/Dockerfile
DamianEdwards: > Does YARP support a simple proxy config, where I can intercept /api HTTP requests, and redirect them to the configured weather API url?

Yes: https://microsoft.github.io/reverse-proxy/articles/config-files.html

Perhaps we could add a YARP project to the solution and use it as the host for all the JS apps. [In theory](https://microsoft.github.io/reverse-proxy/articles/getting-started.html), the YARP project can be as simple as:

```csharp
var builder = WebApplication.CreateBuilder(args);
builder.Services.AddReverseProxy()
    .LoadFromConfig(builder.Configuration.GetSection("ReverseProxy"));
var app = builder.Build();
app.MapReverseProxy();
app.Run();
```

samples/AspireWithJavaScript/AspireJavaScript.React/webpack.config.js
DamianEdwards: Or was that a Vite-only thing?

DamianEdwards: Could we make this auto-include any environmental variables with that prefix perhaps? There really isn't any point using the custom environment variable names if we just have to map them again in the webpack file. Might as well just keep the default names that Service Discovery injects and map those.

IEvangelist: The `REACT_APP_` stuff was from the `create-react-app` template that we removed here. With webpack, nothing seems to happen implicitly. I renamed them, putting them into a `variables` object as to avoid confusion with `process.env`.

DamianEdwards (approved):
Thank you!

PR #233, opened 2024-04-24
Proxy all the things locally. And sit behind nginx, for now.

IEvangelist:
Now React and Vue site behind nginx. Also, trying to use the `DockerBuildArg`- but realizing that really the env var are still what was used by the `nginx` template. 🤓

The resulting containers are a fraction the size of the original. This work sets us up for later pivoting to using YARP.

**Update**

I believe that `azd` doesn't correctly handle the `--build-arg`, the value is empty during the build step. It should be pulling from the `services__weatherapi__https__0` env var.


DamianEdwards:
@IEvangelist can you add some screenshots/video of it running locally for you, and for bonus points do a deployment to ACA via AZD and put the URLs here so we can see it working there too? Thanks!

IEvangelist:
> @IEvangelist can you add some screenshots/video of it running locally for you, and for bonus points do a deployment to ACA via AZD and put the URLs here so we can see it working there too? Thanks!

@DamianEdwards - sure thing:

https://github.com/dotnet/aspire-samples/assets/7679720/e7e7e4b3-64fa-46e6-8941-d88ccc5ff145

Links to deployed versions using `azd`:

- https://angular.yellowtree-fed0ffc7.centralus.azurecontainerapps.io/
- https://react.yellowtree-fed0ffc7.centralus.azurecontainerapps.io/
- https://vue.yellowtree-fed0ffc7.centralus.azurecontainerapps.io/

DamianEdwards (approved):

PR #235, opened 2024-04-26
Remove extension helpers, instead use parameters for passwords.

IEvangelist:
Remove extension helpers, instead use parameters:

- Related to https://github.com/dotnet/docs-aspire/issues/672

### DIFF
--- a/samples/VolumeMount/VolumeMount.AppHost/Program.cs
+++ b/samples/VolumeMount/VolumeMount.AppHost/Program.cs
@@ -2,17 +2,13 @@
 
 // Using a persistent volume mount requires a stable password rather than the default generated one.
 
-var sqlPassword = builder.AddParameter("sqlpassword", secret: true);
-
 // To have a persistent volume across container instances, it must be named.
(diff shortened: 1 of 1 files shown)


alexwolfmsft (approved):

samples/VolumeMount/VolumeMount.AppHost/Program.cs
```diff
@@ -2,13 +2,19 @@
 
 // Using a persistent volume mount requires a stable password rather than the default generated one.
 
+var sqlPassword = builder.AddParameter("sqlpassword", secret: true);
```
davidfowl, line 5: You don't need this either.

```suggestion
```

IEvangelist: Ah, ok - got it! Thanks!

IEvangelist:
Discussed this offline with @davidfowl - closing this in favor of an alternative approach to update the docs, with local-code snippets that tell the story a bit more precisely.

PR #247, opened 2024-04-30
Replace calls to PokeAPI with nuget.org in Standalone dashboard sample

DamianEdwards:
Calling out to public HTTP endpoints not owned by Microsoft isn't something we want to do in our samples, so this changes the Standalone dashboard sample to instead make calls to nuget.orgj


joperezr (approved):
LGTM, thanks for fixing.

blowdart (approved):

PR #251, opened 2024-05-02
Increase test HttpClient retries & optionally wait for resources to transition to target state

DamianEdwards:
Would appreciate your eyes @davidfowl @radical @ReubenBond 

The DatabaseMigrations sample test was flakey as the migrations project is just a console app so has no healthcheck endpoint that can be monitored before trying to call the actual app endpoint, so it would sometimes fail because the retries for the app endpoint would get exhausted before the migration had finished. Now, the test can wait for a specific resource to enter a specific state before continuing, which I use to wait for the migration app to exit before hitting the actual app endpoints.

### DIFF
--- a/tests/SamplesIntegrationTests/AppHostTests.cs
+++ b/tests/SamplesIntegrationTests/AppHostTests.cs
@@ -3,8 +3,8 @@
 
 using System.Net;
 using System.Text.Json;
-using Aspire.Hosting.Dapr;
 using Microsoft.Extensions.DependencyInjection;
+using SamplesIntegrationTests.Infrastructure;
 using Xunit.Abstractions;
 
 namespace SamplesIntegrationTests;
@@ -66,7 +66,7 @@ public class AppHostTests(ITestOutputHelper testOutput)
                     {
                         resilience.TotalRequestTimeout.Timeout = TimeSpan.FromSeconds(300);
                         resilience.AttemptTimeout.Timeout = TimeSpan.FromSeconds(60);
-                        resilience.Retry.MaxRetryAttempts = 5;
+                        resilience.Retry.MaxRetryAttempts = 30;
                         resilience.CircuitBreaker.SamplingDuration = TimeSpan.FromSeconds(300);
                     });
             });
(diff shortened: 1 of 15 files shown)


tests/SamplesIntegrationTests/Infrastructure/ResourceWatcher.cs
```diff
@@ -54,6 +57,14 @@ public override void Dispose()
 
     public Task WaitForResourcesToStop() => _resourcesStoppedTcs.Task;
 
+    public Task WaitForResource(string resourceName, string targetState = "Running", CancellationToken cancellationToken = default)
```
ReubenBond, line 60: We could consider including a provision here to cancel waiters when the ResourceWatcher receives StopAsync (i.e, the waiter will never otherwise be resolved). It's not crucial, though.

ReubenBond (approved):

ReubenBond:
We should consider integrating this functionality back into the core product

tests/SamplesIntegrationTests/Infrastructure/ResourceWatcher.cs
DamianEdwards: Good point, updated.

DamianEdwards: LMK what you think of how I did it.

tests/SamplesIntegrationTests/AppHostTests.cs
```diff
@@ -46,6 +46,16 @@ public async Task TestEndpointsReturnOk(TestEndpoints testEndpoints)
 
         await app.StartAsync(waitForResourcesToStart: true);
 
+        if (testEndpoints.WaitForResources?.Count > 0)
+        {
+            // Wait until each resource transitions to the required state
+            var timeout = TimeSpan.FromMinutes(5);
+            foreach (var (ResourceName, TargetState) in testEndpoints.WaitForResources)
+            {
+                await app.WaitForResource(ResourceName, TargetState, new CancellationTokenSource(timeout).Token);
```
radical, line 55: Would there be any benefit in starting off all the tasks, and then `WaitAll()/WaitAny()` them? For one, we would find out earlier about a task failure which we are not waiting on yet. For example, if we are waiting on the first one, and the 3rd one fails so `WaitForResource` on that would throw - it could be handled sooner.

tests/SamplesIntegrationTests/Infrastructure/DistributedApplicationExtensions.cs
```diff
@@ -160,6 +159,12 @@ public static async Task StartAsync(this DistributedApplication app, bool waitFo
         await resourcesStartingTask;
     }
 
+    public static async Task WaitForResource(this DistributedApplication app, string resourceName, string state = "Running", CancellationToken cancellationToken = default)
```
radical, line 162: This doesn't need to be `async`, and can just return the task.

tests/SamplesIntegrationTests/AppHostTests.cs
DamianEdwards: I considered it briefly but didn't bother 'cos there's only a single test that uses this and it only specifies a single resource, but in general it would make sense due to the reasons you point out.

DamianEdwards: Thinking more about this I actually wanted to preserve order or waiting. That way a theory can define fully the order it wants to wait for resources to move to specific states before it moves on. It's a way of creating order out of the randomness of the fully async/parallel nature of resource orchestration.

PR #289, opened 2024-05-23
Fix incorrect execution strategy name

hognevevle:


AndriySvyryd:
The name is already correct. Though we might consider renaming it to something less redundant, @DamianEdwards

hognevevle:
@AndriySvyryd That's weird, because only after my proposed change does the code compile.
CleanShot 2024-06-24 at 22 12 16

With the original code:
CleanShot 2024-06-24 at 22 13 05

AndriySvyryd:
It's linked in the project https://github.com/hognevevle/aspire-samples/blob/main/samples/DatabaseMigrations/DatabaseMigrations.MigrationService/DatabaseMigrations.MigrationService.csproj#L12
Perhaps an IDE issue?

hognevevle:
Sorry, of course you are right, I didn't realize there was [an override in play here](https://github.com/hognevevle/aspire-samples/blob/main/samples/Shared/RetryingSqlServerRetryingExecutionStrategy.cs). It's was my own code not compiling, not the sample repo, so I was of course missing that file. 🫣

Issue #74, opened 2023-12-27
Add sample to use Aspire with CosmosDB


Kralizek:
I would appreciate if there was a sample showing how to use CosmosDB in Aspire since this simple setup isn't working and I have no clue on what's missing.

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var db = builder.AddAzureCosmosDB("db");

var api = builder.AddProject<Projects.API>("api")
    .WithReference(db);

builder.Build().Run();
```

See: https://github.com/dotnet/aspire/issues/1502

DamianEdwards:
Thanks for the suggestion. @bradygaster mind taking this one? Or perhaps @Pilchie?

Pilchie:
I want to do this, but need to be honest that it likely won't be this month.  In the meantime, if you check out some of the earlier commits of https://github.com/dotnet/aspire/pull/359, it did have some apps in the main app.  For example: https://github.com/dotnet/aspire/pull/359/commits/4f01310c85336c8595a7314e4f0e1ef255148796

bradygaster:
I can do this. I'll have it by end of next week, apologies for the delay.

Pilchie:
You're the best @bradygaster !

chrisrlewis:
> I can do this. I'll have it by end of next week, apologies for the delay.

@bradygaster I'm also looking forward to this, as currently struggling with starting with Cosmos db + Aspire. If the example could include the various ways to connect and also include the required config file settings that would be perfect.

DamianEdwards:
@bradygaster you still planning on doing this one?

bradygaster:
I can do it if we didn't get one done. I started adding it to an app but got side-tracked.

bradygaster:
Update to @DamianEdwards and @Pilchie and @markjbrown that I've created the sample with some help from the nice folks in Cosmos DB. 

Let me know (in the form of a pull request or issue) if there's anything else y'all think this needs, or any tweaks, and I'll facilitate. 

Also to @chrisrlewis - not sure if this satisfies your desire for "all the ways to connect," but I wanted to share this with you to see if this helps at all.

Pilchie:
LGTM.  Thanks @bradygaster.

markjbrown:
Love it. Thanks @bradygaster !!!

Kralizek:
would it be possible to expand the example with the use of the local emulator?

PR #321, opened 2024-06-13
added cosmosdb sample

bradygaster:
Adds Cosmos DB sample per #74


samples/AspireWithCosmosDb/README.md
```diff
@@ -0,0 +1,59 @@
...
+- dotnet-aspire
+page_type: sample
+name: ".NET Aspire with Azure Cosmos DB"
+urlFragment: "aspire-cosmosdb"
+description: "A sample todo app demonstrating how .NET Aspire apps can use Azure Cosmos DB."
+---
+
+# Cosmic Aspirations - .NET Aspire with Cosmos DB
+
+This is a simple app that shows how to use Cosmos DB to store data from a .NET Aspire app.
```
Pilchie, line 15:
```suggestion
This is a simple app that shows how to use Azure Cosmos DB to store data from a .NET Aspire app. 
```

samples/AspireWithCosmosDb/README.md
```diff
@@ -0,0 +1,59 @@
...
+## Building the sample
+
+To download and run the sample, follow these steps:
+
+1. Clone the `dotnet/aspire-samples` repository.
+
+2. In Visual Studio (2022 or later):
+    1. On the menu bar, choose **File** > **Open** > **Project/Solution**.
+    2. Navigate to the folder that holds the sample code, and open the `AspireWithCosmos.sln` solution (`.sln`) file.
+    3. Right click the `AspireWithCosmos.AppHost` project in the solution explore and choose it as the startup project.
```
Pilchie, line 36: You can make this the default by having it be the first project listed in the `.sln` file.

samples/AspireWithCosmosDb/README.md
```diff
@@ -0,0 +1,59 @@
...
+    4. Right click the Connected Services node in the `AspireWithCosmos.AppHost` project and select Azure Resource Provisioning Settings to set up your project for Azure resource auto-provisioning. 
+    
+       ![Configuring the app in Visual Studio](images/image-20240613155416498.png)
+    
+    5. Choose the <kbd>F5</kbd> key to run with debugging, or <kbd>Ctrl</kbd>+<kbd>F5</kbd> keys to run the project without debugging.
+    
+3. From the command line:
+   1. Navigate to the folder that holds the sample code.
+   
+   2. At the command line, type [`dotnet run --project .\AspireWithCosmos.AppHost\AspireWithCosmos.AppHost.csproj`](https://docs.microsoft.com/dotnet/core/tools/dotnet-run).
```
Pilchie, line 46: This won't provision a cosmos DB account, right?

samples/AspireWithCosmosDb/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+1. Clone the `dotnet/aspire-samples` repository.
+
+2. In Visual Studio (2022 or later):
+    1. On the menu bar, choose **File** > **Open** > **Project/Solution**.
+    2. Navigate to the folder that holds the sample code, and open the `AspireWithCosmos.sln` solution (`.sln`) file.
+    3. Right click the `AspireWithCosmos.AppHost` project in the solution explore and choose it as the startup project.
+    4. Right click the Connected Services node in the `AspireWithCosmos.AppHost` project and select Azure Resource Provisioning Settings to set up your project for Azure resource auto-provisioning. 
+    
+       ![Configuring the app in Visual Studio](images/image-20240613155416498.png)
```
Pilchie, line 39: I think it's typical to do VS screenshots with the default theme, since it has accessible contrast ratios.

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+public class DatabaseBootstrapper(CosmosClient cosmosClient) : IHostedService
+{
+    public async Task StartAsync(CancellationToken cancellationToken)
+    {
+        await cosmosClient.CreateDatabaseIfNotExistsAsync("tododb");
+        var database = cosmosClient.GetDatabase("tododb");
+        await database.CreateContainerIfNotExistsAsync(new ContainerProperties("todos", Todo.UserIdPartitionKey));
+    }
+
+    public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
```
Pilchie, line 63: Consider `Dispose` on your `cosmosClient` here.

Pilchie (approved):

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/Program.cs
davidfowl: If it's in the DI container, and disposable, it'll get disposed.

samples/AspireWithCosmosDb/README.md
bradygaster: Yes, auto-provisioning will.

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/Program.cs
```diff
@@ -0,0 +1,78 @@
...
+
+// Convenience class for reusing boilerplate code
+public static class CosmosClientTodoAppExtensions
+{
+    public static Container GetAppDataContainer(this CosmosClient cosmosClient)
+    {
+        var database = cosmosClient.GetDatabase("tododb");
+        var todos = database.GetContainer("todos");
+
+        if(todos == null) throw new ApplicationException("Cosmos DB collection missing.");
```
DamianEdwards, line 74:
```suggestion
        var todos = database.GetContainer("todos") ?? throw new ApplicationException("Cosmos DB collection missing.");
```

samples/AspireWithCosmosDb/AspireWithCosmos.AppHost/AspireWithCosmos.AppHost.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <IsAspireHost>true</IsAspireHost>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\AspireWithCosmos.ApiService\AspireWithCosmos.ApiService.csproj" />
+    <ProjectReference Include="..\AspireWithCosmos.Web\AspireWithCosmos.Web.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Aspire.Hosting.AppHost" Version="8.0.0" />
```
DamianEdwards, line 17:
```suggestion
    <PackageReference Include="Aspire.Hosting.AppHost" Version="8.0.1" />
```

samples/AspireWithCosmosDb/AspireWithCosmos.AppHost/Program.cs
```diff
@@ -0,0 +1,12 @@
+var builder = DistributedApplication.CreateBuilder(args);
+
+var cosmos = builder.AddAzureCosmosDB("cosmos");
+
+var apiService = builder.AddProject<Projects.AspireWithCosmos_ApiService>("apiservice")
+                        .WithReference(cosmos);
+
+builder.AddProject<Projects.AspireWithCosmos_Web>("webfrontend")
+    .WithExternalHttpEndpoints()
+    .WithReference(apiService);
```
DamianEdwards, line 10: nit: Be consistent with your indentation style, either single indent or aligned with statement above.

samples/AspireWithCosmosDb/AspireWithCosmos.ServiceDefaults/AspireWithCosmos.ServiceDefaults.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsAspireSharedProject>true</IsAspireSharedProject>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+
+    <PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="8.3.0" />
+    <PackageReference Include="Microsoft.Extensions.ServiceDiscovery" Version="8.0.0" />
```
DamianEdwards, line 14:
```suggestion
    <PackageReference Include="Microsoft.Extensions.ServiceDiscovery" Version="8.0.1" />
```

samples/AspireWithCosmosDb/AspireWithCosmos.ServiceDefaults/AspireWithCosmos.ServiceDefaults.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsAspireSharedProject>true</IsAspireSharedProject>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+
+    <PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="8.3.0" />
```
DamianEdwards, line 13:
```suggestion
    <PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="8.6.0" />
```

samples/AspireWithCosmosDb/AspireWithCosmos.ServiceDefaults/AspireWithCosmos.ServiceDefaults.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <PackageReference Include="Microsoft.Extensions.ServiceDiscovery" Version="8.0.0" />
+    <PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.8.1" />
+    <PackageReference Include="OpenTelemetry.Extensions.Hosting" Version="1.8.1" />
+    <PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.8.1" />
+    <PackageReference Include="OpenTelemetry.Instrumentation.Http" Version="1.8.1" />
+    <PackageReference Include="OpenTelemetry.Instrumentation.Runtime" Version="1.8.0" />
```
DamianEdwards, line 19: Check all of these for updates.

samples/AspireWithCosmosDb/AspireWithCosmos.ServiceDefaults/Extensions.cs
```diff
@@ -0,0 +1,111 @@
...
+            {
+                metrics.AddAspNetCoreInstrumentation()
+                    .AddHttpClientInstrumentation()
+                    .AddRuntimeInstrumentation();
+            })
+            .WithTracing(tracing =>
+            {
+                tracing.AddAspNetCoreInstrumentation()
+                    // Uncomment the following line to enable gRPC instrumentation (requires the OpenTelemetry.Instrumentation.GrpcNetClient package)
+                    //.AddGrpcClientInstrumentation()
```
DamianEdwards, line 56: Remove lines not needed in this specific sample.

samples/AspireWithCosmosDb/AspireWithCosmos.ServiceDefaults/Extensions.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            builder.Services.AddOpenTelemetry().UseOtlpExporter();
+        }
+
+        // Uncomment the following lines to enable the Azure Monitor exporter (requires the Azure.Monitor.OpenTelemetry.AspNetCore package)
+        //if (!string.IsNullOrEmpty(builder.Configuration["APPLICATIONINSIGHTS_CONNECTION_STRING"]))
+        //{
+        //    builder.Services.AddOpenTelemetry()
+        //       .UseAzureMonitor();
+        //}
```
DamianEdwards, line 79: Remove lines not needed in this specific sample.

samples/AspireWithCosmosDb/AspireWithCosmos.Tests/AspireWithCosmos.Tests.csproj
```diff
@@ -0,0 +1,28 @@
...
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
+    <IsTestProject>true</IsTestProject>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Aspire.Hosting.Testing" Version="8.0.0" />
```
DamianEdwards, line 12:
```suggestion
    <PackageReference Include="Aspire.Hosting.Testing" Version="8.0.1" />
```

samples/AspireWithCosmosDb/AspireWithCosmos.Tests/AspireWithCosmos.Tests.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <PackageReference Include="coverlet.collector" Version="6.0.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.9.0" />
+    <PackageReference Include="xunit" Version="2.5.3" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.5.3" />
```
DamianEdwards, line 16: Check these for updates against the other samples in the repo.

samples/AspireWithCosmosDb/README.md
Pilchie: Right, but the dark theme was excused from meeting accessibility requirements.

samples/AspireWithCosmosDb/AspireWithCosmos.Tests/WebTests.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace AspireWithCosmos.Tests;
+
+public class WebTests
+{
+    [Fact]
+    public async Task GetWebResourceRootReturnsOkStatusCode()
+    {
+        // Arrange
+        var appHost = await DistributedApplicationTestingBuilder.CreateAsync<Projects.AspireWithCosmos_AppHost>();
+        await using var app = await appHost.BuildAsync();
```
DamianEdwards, line 12: You'll want to add the standard HTTP resiliency handler here for this to be reliable but given this just looks like the standard test template I'd question if it has value in this sample so perhaps just remove it.

samples/AspireWithCosmosDb/README.md
Pilchie: At the command line, what about people that don't have VS?

samples/AspireWithCosmosDb/AspireWithCosmos.Web/TodoApiClient.cs
```diff
@@ -0,0 +1,42 @@
...
+        => await RestService.For<ITodoApiClient>(_httpClient).Update(id, todo);
+
+    public async Task Delete(string userId, string id)
+        => await RestService.For<ITodoApiClient>(_httpClient).Delete(userId, id);
+}
+
+// The Todo service model used for transmitting data
+public record Todo(string Description, string UserId)
+{
+    public string id { get; set; }
```
DamianEdwards, line 40:
```suggestion
    public required string id { get; set; }
```

samples/AspireWithCosmosDb/images/image-20240613154158603.png
DamianEdwards, line 1: Let's use a cleaner, more default looking terminal for screenshots.

DamianEdwards:
@bradygaster you'll need to figure out how to make this sample pass the test cleanly as currently the apiservice fails to start it seems.

samples/AspireWithCosmosDb/AspireWithCosmos.AppHost/Program.cs
```diff
@@ -0,0 +1,12 @@
+var builder = DistributedApplication.CreateBuilder(args);
+
+var cosmos = builder.AddAzureCosmosDB("cosmos");
```
DamianEdwards, line 3: Do we want to make this use the emulator by default so it works straight out of the box? Perhaps add a comment explaining the options?

samples/AspireWithCosmosDb/README.md
bradygaster: Either way it will. If you don't have VS the logs will walk you through what to configure.

samples/AspireWithCosmosDb/AspireWithCosmos.AppHost/Program.cs
bradygaster: I don't know if I manually changed this. But yes. Agreed.

samples/AspireWithCosmosDb/AspireWithCosmos.ServiceDefaults/Extensions.cs
bradygaster: I'd like to leave that in so folks who need the guidance have it in this sample as well as in the default template content. Given it's an Azure Cosmos DB sample, i presume the chances are good folks will want to add this so this guidance seems appropriate to leave in as a comment, like your guidance on adding comments for using the emulator in the app host.

samples/AspireWithCosmosDb/AspireWithCosmos.Tests/WebTests.cs
bradygaster: i was contemplating adding playwright tests here to pump through the ui but i think that'd conflate the sample.

bradygaster:
Pretty sure I captured all the changes suggested by @DamianEdwards and @Pilchie save 1 - i couldn't get the app to work properly when i have `RunAsEmulator` in the App Host code. If @Pilchie has guidance on how to work past the error i've sent him i'll implement it, otherwise i think this has everything everyone requested and is ready to go.

DamianEdwards:
I still see some build warnings and the test still isn't passing?

bradygaster:
added the `RunAsEmulator` command as it matches the implementation in the playground project and the comment @DamianEdwards requested, in lieu of the fact that i couldn't get it working with the emulator.

bradygaster:
Guess the emulator is broken on my machine. For once, it not working on my machine isn't a bad thing? :)

DamianEdwards:
Only thing left would be to add a test case at https://github.com/dotnet/aspire-samples/blob/main/tests/SamplesIntegrationTests/AppHostTests.cs#L125 so that we get proper coverage of this new sample.

bradygaster:
tests seem to be passing now - let me know if there's anything else this one needs, folks.

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/Program.cs
```diff
@@ -51,16 +61,59 @@ public record Todo(string Description, string id, string UserId, bool IsComplete
...
+            : _dbCreationFailed
+                ? HealthCheckResult.Unhealthy("Database creation failed.")
+                : HealthCheckResult.Degraded("Database creation is still in progress.");
+        return Task.FromResult(status);
     }
 
-    public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        // The Cosmos DB emulator can take a very long time to start (multiple minutes) so use a custom resilience strategy
```
bradygaster, line 81: cc @Pilchie so he sees how @DamianEdwards was able to work through this - I pinged the team today so mention we'd run into these issues, so wanted to give y'all an update on the completion. Thanks @DamianEdwards!

DamianEdwards:
@bradygaster @Pilchie I pushed some changes that hopefully make this more reliable:

- Added retries to the `DatabaseBootstrapper` using a Polly resiliency pipeline with enough retries and time to allow the emulator container to start (hopefully)
- Changed the `DatabaseBootstrapper` from a raw `IHostedService` to a `BackgroundService` so its startup doesn't impact the lifecycle of the app. As it was, Kestrel couldn't start until the `DatabaseBootstrapper` had successfully completed, and if it failed the app would just crash.
- Made `DatabaseBootstrapper` double as an `IHealthCheck` so it can report its status and the test can ensure /health succeeds before moving on to trying to hit /todos

It seems problematic that the Cosmos DB component doesn't enable enough retries by default to allow the emulator to startup. Also I found it confusing that the Cosmos DB resource reports itself as having only a TCP endpoint but then it clearly uses HTTP at runtime, but because it doesn't use `IHttpClientFactory` it doesn't get any of the default Aspire HTTP resiliency applied. I attempted to wire the Cosmos client up to `IHttpClientFactory` but found it more complex than simply wrapping retries around the creation logic due to having to separate it from the default resiliency (as that wasn't resilient **enough**) and having to manually handle the HTTPS cert again.

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/Program.cs
davidfowl: We should put this code in the component.

cc @eerhardt

eerhardt: 1. @Pilchie mentioned offline that they were going to change the emulator to not respond until it was ready. That would solve the issues with getting SSL errors during startup.
2. We could add this to the HttpFactory by default. It would be better if the library had resilience built into it though.

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/AspireWithCosmos.ApiService.csproj
```diff
@@ -0,0 +1,21 @@
...
+	<PropertyGroup>
+		<TargetFramework>net8.0</TargetFramework>
+		<ImplicitUsings>enable</ImplicitUsings>
+		<Nullable>enable</Nullable>
+	</PropertyGroup>
+
+	<ItemGroup>
+		<PackageReference Include="Aspire.Microsoft.Azure.Cosmos" Version="8.0.1" />
+		<PackageReference Include="Azure.Identity" Version="1.12.0" />
+		<PackageReference Include="Microsoft.Azure.Cosmos" Version="3.41.0" />
```
eerhardt, line 12:
```suggestion
```
Why are these necessary? You should be able to just have line 10.

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/Program.cs
```diff
@@ -0,0 +1,140 @@
...
+
+builder.AddServiceDefaults();
+builder.Services.AddProblemDetails();
+builder.Services.AddEndpointsApiExplorer();
+builder.Services.AddOpenApiDocument();
+
+builder.AddAzureCosmosClient("cosmos");
+builder.Services.AddSingleton<DatabaseBootstrapper>();
+builder.Services.AddHealthChecks()
+    .Add(new("cosmos", sp => sp.GetRequiredService<DatabaseBootstrapper>(), null, null));
```
eerhardt, line 16: This is odd to have in our sample, when the decision was to not have a health check in the component.

https://github.com/dotnet/aspire/pull/359#discussion_r1363061960

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/Program.cs
DamianEdwards: I think if we set the HttpClientFactory option we then own all other aspects setting up the HttpClient instance that the Cosmos client currently handles, including those affected by properties on the connecting string, e.g. disabling the SSL validation check. This is what I ran into when attempting that approach.

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/AspireWithCosmos.ApiService.csproj
DamianEdwards: Vulnerability in Azure.Identity in versions lower than that and we haven't updated the component yet.

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/Program.cs
DamianEdwards: Right but this isn't a health check of Cosmos, it's a health check of whether the bootstrapper has completed successfully. Once the bootstrapper has completed this health check is always healthy and doesn't hit Cosmos on its own.

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/Program.cs
bradygaster: Do we want to merge this as-is or edit the code in the component in such a manner we can make the sample less-meh? I vote for merging it and then I'll update it once we change the component if @eerhardt informs me once the work is done. :)

eerhardt: I'd agree we should wait for the component changes.

FYI - the component is owned by @Pilchie's team. We should work with them to get it updated as necessary.

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/AspireWithCosmos.ApiService.csproj
DamianEdwards: @bradygaster @eerhardt We shouldn't have samples with vulnerable packages, even transitively. This is what I see in VS for the transitive reference to Azure.Identity 
image

Customers need to update transitive references to resolve these issues and so IMO it's appropriate to do so in samples too. When we ship updated versions of the Aspire packages that update the transitive versions used we can remove the explicit references.

samples/AspireWithCosmosDb/AspireWithCosmos.ApiService/Program.cs
Pilchie: Just wanted to comment that I'm aware of this and following up.  I think there is a plan to build some of this into the container itself so that this sort of thing isn't necessary, but I will let you know.  cc @kirankumarkolli

PR #327, opened 2024-06-17
Add ability for ContainerResource to reach ProjectResource by name in Run mode

dombrovsky:
This modifies Metrics sample.

The goal is to change in `prometheus.yml` this
`- targets: ['host.docker.internal:5048'] # hard-coded port matches launchSettings.json`
to this
`- targets: ['app']`,
so that prometheus container can reach metricsapp that runs on host in development environment by resource name instead of hardcoded endpoint.

In this PR this is achieved by:
1. Creating custom docker network and adding all ContainerResource to it.
2. Adding simple Nginx container with network alias equal to ProjectResource name which should be accessible by container.


dombrovsky:
@dotnet-policy-service agree

davidfowl:
I don't think we should make the metrics example more complex.

davidfowl:
This container proxy resource is a very cool stand alone sample though.

dombrovsky:
Thanks for review. I'll prepare it as a standalone sample and update PR.

PR #362, opened 2024-07-15
.NET Aspire 8.1 Samples Updates

mitchdenny:
This PR is gathering up all the .NET Aspire 8.1 sample updates into a single PR so we can merge them as a batch.


NuGet.config
```diff
@@ -3,12 +3,16 @@
   <packageSources>
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="dotnet8feed" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" protocolVersion="3" />
```
danmoseley, line 6: We presumably want to remove this as soon as 8.1 is out.

danmoseley (approved):
This assumes we want this in main. I see 3 options
1. just validate in the PR, and only merge when 8.1 goes out (but removing the nuget.config line). this is what we've done in eshop, eg.. 
2. merge into main, remove the nuget.config line when 8.1 goes out. downside: most people consuming aspire probably want to start with stable builds.
3. merge into release/v.next type branch. not sure the value of this and it would need infra setting up.

I'm assuming you're planning # 1?

NuGet.config
mitchdenny: Yes we will remove this once 8.1 ships and we merge this PR (as well as updating all the package versions).

danmoseley:
> Please how can ii contribute

Hi @dom3691 welcome. What kind of contribution are you looking to make? There are some open issues in this repo you can look at. Perhaps you might want to look at Aspire proper -- see https://github.com/dotnet/aspire/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22 and find one that interests you, then make a comment on that.

danmoseley:
eg., this one looks like a potentially good first issue
https://github.com/dotnet/aspire/issues/2649
if you did this, I would make a single change (just one fix in one place) and put up a PR to check you're doing the kind of thing expected, before doing more.

mitchdenny:
> I'm assuming you're planning # 1?

Yes. Thats my plan.

.gitattributes
```diff
@@ -3,6 +3,8 @@
 ###############################################################################
 * text=auto
 *.sh text eol=lf
+*.mod binary
```
eerhardt, line 6: @mitchdenny - what is this change for?

tests/SamplesIntegrationTests/AppHostTests.cs
```diff
@@ -97,7 +97,7 @@ public async Task TestEndpointsReturnOk(TestEndpoints testEndpoints)
             }
         }
 
-        appHostLogs.EnsureNoErrors();
+        // DCP is now sending stderr?? appHostLogs.EnsureNoErrors();
```
eerhardt, line 100: We should not merge with this change.

eerhardt (dismissed):

.gitattributes
danmoseley: I'm guessing git is mis-detecting those go files as text, and therefore mangling the line endings (which are actually not line endings)

danmoseley: oh, hold on, they are apparently text. which is I guess your point.. Maybe go complains if they have \r\n

mitchdenny: Exactly. It kept getting them wrong.

eerhardt: I'm not sure telling git that these are binary is the right call.

Looking at other .gitattributes:

https://grep.app/search?q=%2A.mod&words=true&filter[lang][0]=Git%20Attributes

People mostly tell it that `eol=lf`.

radical:
Using [linux-dev-certs](https://github.com/tmds/linux-dev-certs) tool seems to fix the failures.

tests/SamplesIntegrationTests/AppHostTests.cs
eerhardt: This has been resolved in the latest change.

mitchdenny:
Merging thanks for resolving this @eerhardt

PR #380, opened 2024-07-23
AddPythonProject sample

mitchdenny:
Builds on #362 but adds Python. Not included in the main PR because I don't want to block that if we have dependency issues with Python.


.gitignore
```diff
@@ -431,3 +431,2037 @@ FodyWeavers.xsd
 **/*/azure.yaml
 **/*/next-steps.md
 
+samples/AspireWithPython/InstrumentedPythonProject/.venv/pyvenv.cfg
```
IEvangelist, line 434: Yuck, just ignore `**/*/.venv/` and be done with it. No reason individually ignore all these files.

mitchdenny: OMG just right clicked on the folder and selected ignore.

samples/AspireWithPython/AspireWithPython.AppHost/Program.cs
```diff
@@ -0,0 +1,6 @@
+﻿var builder = DistributedApplication.CreateBuilder(args);
+
+builder.AddPythonProject("instrumented-python-project", "..\\InstrumentedPythonProject", "main.py")
```
IEvangelist, line 3: NIT: .NET supports both backslash and forward slash, I prefer to standardize on forward slash as it's less characters to type and it doesn't require an escape sequence.

```suggestion
builder.AddPythonProject("instrumented-python-project", "../InstrumentedPythonProject", "main.py")
```

IEvangelist (approved):

mitchdenny:
Merging on red as build failures seem unrelated.

PR #383, opened 2024-07-24
Workaround sqlserver 2022-latest latest image failing

radical:
.. because the path `/opt/mssql-tools/bin/sqlcmd`
changed to `/opt/mssql-tools18/bin/sqlcmd`. And this is being used in
`samples/DatabaseContainers/DatabaseContainers.AppHost/sqlserverconfig/configure-db.sh`.

https://github.com/microsoft/mssql-docker/issues/892

Also, adds `-C` to trust the server certificate (see https://learn.microsoft.com/en-us/sql/tools/sqlcmd/sqlcmd-utility?view=sql-server-ver16&tabs=go%2Cwindows&pivots=cs1-bash#-c).


eerhardt (approved):
Thanks!

PR #455, opened 2024-09-05
Simple AzureOpenAI sample

samsp-msft:
Simple sample for using the OpenAI components


samples/AzureAI/AzureAISample.AppHost/appsettings.Development.json
```diff
@@ -0,0 +1,14 @@
+{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft.AspNetCore": "Warning"
+    }
+  },
+  "Azure": {
```
jmatthiesen, line 8: You'll want to remove this from the committed file since it's specific to your environment/setup.

PR #525, opened 2024-10-28
Update EF migrations to match EF tools version used

DamianEdwards:
Recreates all the migrations so they match the version of EF being used.


AndriySvyryd (approved):
Looks like these were generated using 8.0.10.
Did you verify that there is no longer an exception thrown when using EF 9 with these?

DamianEdwards:
> Did you verify that there is no longer an exception thrown when using EF 9 with these?

Going to try that next.

PR #529, opened 2024-10-31
Update the Node sample to use the dev cert

DamianEdwards:
Updates the Node.js sample to use the ASP.NET Core HTTPS dev certificate.

@mitchdenny @davidfowl you might want to take a look at the use of eventing here. I had some issues trying to achieve the method layering with anything other than the `BeforeStartEvent` but @davidfowl pointed out that work in that event blocks the dashboard from starting which is unfortunate but ultimately probably not a huge concern in this case. Specifically the need here is to do work (exporting the dev cert) that if successful results in changes to the resource (adding env vars, endpoints, bind mounts, and maybe more env vars that reference the added endpoints), and has access to a resource logger during that work.

### DIFF
--- a/samples/Shared/DevCertHostingExtensions.cs
+++ b/samples/Shared/DevCertHostingExtensions.cs
@@ -33,6 +33,7 @@ public static class DevCertHostingExtensions
                 if (!exported)
                 {
                     // The export failed for some reason, don't configure the resource to use the certificate.
+                    return;
                 }
 
                 if (builder.Resource is ContainerResource containerResource)
--- a/samples/Shared/DevCertHostingExtensions.cs
+++ b/samples/Shared/DevCertHostingExtensions.cs
@@ -75,7 +75,7 @@ public static class DevCertHostingExtensions
     {
         // Exports the ASP.NET Core HTTPS development certificate & private key to PEM files using 'dotnet dev-certs https' to a temporary
         // directory and returns the path.
-        // TODO: Check if we're running on a platform that already has the cert and key exported to a file (e.g. macOS) and just use those intead.
+        // TODO: Check if we're running on a platform that already has the cert and key exported to a file (e.g. macOS) and just use those instead.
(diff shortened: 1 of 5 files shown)


samples/Shared/DevCertHostingExtensions.cs
```diff
@@ -0,0 +1,184 @@
...
+            {
+                var logger = e.Services.GetRequiredService<ResourceLoggerService>().GetLogger(builder.Resource);
+
+                // Export the ASP.NET Core HTTPS development certificate & private key to files and configure the resource to use them via
+                // the specified environment variables.
+                var (exported, certPath, certKeyPath) = await TryExportDevCertificateAsync(builder.ApplicationBuilder, logger);
+
+                if (!exported)
+                {
+                    // The export failed for some reason, don't configure the resource to use the certificate.
```
mitchdenny, line 35: return?

mitchdenny:
I think this is fine, after all pre-eventing we would have used the `BeforeStartAsync` method on the lifecycle hook to do it.

I am curious why `BeforeResourceStartEvent` didn't work for you. It runs synchronously and blocks on a per resource basis, but ultimately so does `BeforeStartEvent`.

samples/Shared/DevCertHostingExtensions.cs
DamianEdwards: 🤦‍♂️

DamianEdwards:
> I am curious why BeforeResourceStartEvent didn't work for you. It runs synchronously and blocks on a per resource basis, but ultimately so does BeforeStartEvent.

@mitchdenny when I tried to use `BeforeResourceStartEvent` I got errors from the `ReferenceExpression` referencing the endpoint added in the event callback.

mitchdenny:
Ah it's because you are after endpoint allocation which means any endpoint added won't end up with its AllocatedEndpoint property being added so the reference expression will fail.

mitchdenny (approved):

PR #531, opened 2024-11-05
Fix up Azure Functions sample for publishing

captainsafia:
This PR fixes up a couple of issues with the Azure Functions sample that make it unpublishable:

- Enable external HTTP access on the Blazor front-end resource
- Configure the appropriate roles to support Blob triggers on the Azure Functions host
- Allow anonymous access to blobs in the storage account to support UI display in the Blazor front-end
- Reference `SkiaSharp.NativeAssets.Linux.NoDependencies` package to support running SkiaSharp in Linux containers on ACA

### DIFF
--- a/samples/AspireWithAzureFunctions/ImageGallery.AppHost/Program.cs
+++ b/samples/AspireWithAzureFunctions/ImageGallery.AppHost/Program.cs
@@ -10,8 +10,6 @@ var storage = builder.AddAzureStorage("storage").RunAsEmulator()
         // Storage Account Contributor and Storage Blob Data Owner roles are required by the Azure Functions host
         construct.Add(storageAccount.CreateRoleAssignment(StorageBuiltInRole.StorageAccountContributor, construct.PrincipalTypeParameter, construct.PrincipalIdParameter));
         construct.Add(storageAccount.CreateRoleAssignment(StorageBuiltInRole.StorageBlobDataOwner, construct.PrincipalTypeParameter, construct.PrincipalIdParameter));
-        // Anonymous access to Blobs required by the Blazor front-end
-        storageAccount.AllowBlobPublicAccess = true;
     });
 var blobs = storage.AddBlobs("blobs");
 var queues = storage.AddQueues("queues");
--- a/samples/AspireWithAzureFunctions/ImageGallery.AppHost/Program.cs
+++ b/samples/AspireWithAzureFunctions/ImageGallery.AppHost/Program.cs
@@ -10,6 +10,8 @@ var storage = builder.AddAzureStorage("storage").RunAsEmulator()
         // Storage Account Contributor and Storage Blob Data Owner roles are required by the Azure Functions host
         construct.Add(storageAccount.CreateRoleAssignment(StorageBuiltInRole.StorageAccountContributor, construct.PrincipalTypeParameter, construct.PrincipalIdParameter));
         construct.Add(storageAccount.CreateRoleAssignment(StorageBuiltInRole.StorageBlobDataOwner, construct.PrincipalTypeParameter, construct.PrincipalIdParameter));
+        // Ensure that public access to blobs is disabled
+        storageAccount.AllowBlobPublicAccess = false;
     });
 var blobs = storage.AddBlobs("blobs");
 var queues = storage.AddQueues("queues");
--- a/samples/AspireWithAzureFunctions/ImageGallery.AppHost/Program.cs
+++ b/samples/AspireWithAzureFunctions/ImageGallery.AppHost/Program.cs
@@ -1,17 +1,31 @@
-﻿var builder = DistributedApplication.CreateBuilder(args);
+﻿using Azure.Provisioning.Storage;
 
-var storage = builder.AddAzureStorage("storage").RunAsEmulator();
+var builder = DistributedApplication.CreateBuilder(args);
+
+var storage = builder.AddAzureStorage("storage").RunAsEmulator()
+    .ConfigureConstruct((ResourceModuleConstruct construct) =>
+    {
+        var storageAccount = construct.GetResources().OfType<StorageAccount>().FirstOrDefault(r => r.IdentifierName == "storage")
+            ?? throw new InvalidOperationException($"Could not find configured storage account with name 'storage'");
+        // Storage Account Contributor and Storage Blob Data Owner roles are required by the Azure Functions host
+        construct.Add(storageAccount.CreateRoleAssignment(StorageBuiltInRole.StorageAccountContributor, construct.PrincipalTypeParameter, construct.PrincipalIdParameter));
+        construct.Add(storageAccount.CreateRoleAssignment(StorageBuiltInRole.StorageBlobDataOwner, construct.PrincipalTypeParameter, construct.PrincipalIdParameter));
+        // Anonymous access to Blobs required by the Blazor front-end
+        storageAccount.AllowBlobPublicAccess = true;
+    });
 var blobs = storage.AddBlobs("blobs");
 var queues = storage.AddQueues("queues");
 
 var functions = builder.AddAzureFunctionsProject<Projects.ImageGallery_Functions>("functions")
                        .WithReference(queues)
                        .WithReference(blobs)
-                       .WaitFor(storage);
+                       .WaitFor(storage)
+                       .WithHostStorage(storage);
 
 builder.AddProject<Projects.ImageGallery_FrontEnd>("frontend")
        .WithReference(queues)
        .WithReference(blobs)
-       .WaitFor(functions);
+       .WaitFor(functions)
+       .WithExternalHttpEndpoints();
 
 builder.Build().Run();
--- a/samples/AspireWithAzureFunctions/ImageGallery.FrontEnd/ImageUrl.cs
+++ b/samples/AspireWithAzureFunctions/ImageGallery.FrontEnd/ImageUrl.cs
@@ -1,0 +1,37 @@
+﻿namespace ImageGallery.FrontEnd;
+
+internal static class ImageUrl
+{
+    public static readonly string PathPrefix = "/images";
+    public static readonly string RoutePattern = PathPrefix + "/{slug}";
+
+    public static string GetImageUrl(string slug) => $"{PathPrefix}/{slug}";
+
+    public static string GetThumbnailUrl(string slug) => GetImageUrl(slug) + "?thumbnail=true";
+
+    public static string CreateNameSlug(string name)
+    {
+        var extension = Path.GetExtension(name);
+
+        var chars = name.AsSpan();
+        Span<char> buffer = stackalloc char[chars.Length];
+
+        // Replace invalid characters with '-'
+        // Valid chars are letters, digits, '-', and '_'
+        for (var i = 0; i < chars.Length - extension.Length; i++)
+        {
+            var c = chars[i];
+            buffer[i] = char.IsLetterOrDigit(c) || c == '-' || c == '_' ? c : '-';
+        }
+
+        // Add extension back
+        for (var i = 0; i < extension.Length; i++)
+        {
+            buffer[chars.Length - extension.Length + i] = extension[i];
+        }
+
+        var slug = new string(buffer);
+
+        return slug;
+    }
+}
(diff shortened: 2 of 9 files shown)


IEvangelist (approved):

bradygaster:
cloned and tried deploying - worked like a champ. ship it.

DamianEdwards:
Adding back `WithHostStorage()` is leading to the test issue WRT resource(s) being removed after waiting for them has started. I'll take a look at the test to see if I can refactor it to handle it.

captainsafia:
> Wow, that worked 😮

Should we kick CI a couple of times to sanity check?

DamianEdwards:
> Should we kick CI a couple of times to sanity check?

Yeah I don't even think the logic I added is being hit but instead it's working because now `Start` and `Wait` are awaited sequentially now. I previously had to start them in parallel based on discussions I had with @davidfowl but I can't remember the details and it's seemingly passing now anyway.

samples/AspireWithAzureFunctions/ImageGallery.AppHost/Program.cs
```diff
@@ -1,17 +1,31 @@
...
+var storage = builder.AddAzureStorage("storage").RunAsEmulator()
+    .ConfigureConstruct((ResourceModuleConstruct construct) =>
+    {
+        var storageAccount = construct.GetResources().OfType<StorageAccount>().FirstOrDefault(r => r.IdentifierName == "storage")
+            ?? throw new InvalidOperationException($"Could not find configured storage account with name 'storage'");
+        // Storage Account Contributor and Storage Blob Data Owner roles are required by the Azure Functions host
+        construct.Add(storageAccount.CreateRoleAssignment(StorageBuiltInRole.StorageAccountContributor, construct.PrincipalTypeParameter, construct.PrincipalIdParameter));
+        construct.Add(storageAccount.CreateRoleAssignment(StorageBuiltInRole.StorageBlobDataOwner, construct.PrincipalTypeParameter, construct.PrincipalIdParameter));
+        // Anonymous access to Blobs required by the Blazor front-end
+        storageAccount.AllowBlobPublicAccess = true;
```
davidfowl, line 14: This is interesting..

DamianEdwards: Is it because the Blazor frontend is WebAssembly? If not, we should secure it, no?

DamianEdwards: OK so Blazor frontend is SSR, not WebAssembly. Why can't we secure this access?

DamianEdwards: Is it just so we don't have to have an endpoint in the app that proxies the images from the storage account, but rather we render images with their `src` set directly to the blob?

DamianEdwards: Should be simple enough to do with a Yarp forwarder?

DamianEdwards: Actually this is likely better as a custom endpoint. I'll have a go at adding it.

DamianEdwards: OK added some code, would appreciate eyes on! @captainsafia @bradygaster

samples/AspireWithAzureFunctions/ImageGallery.FrontEnd/ImageUrl.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace ImageGallery.FrontEnd;
+
+internal static class ImageUrl
+{
+    public static readonly string PathPrefix = "/images";
+    public static readonly string RoutePattern = PathPrefix + "/{slug}";
+
+    public static string GetImageUrl(string slug) => $"{PathPrefix}/{slug}";
+
+    public static string GetThumbnailUrl(string slug) => GetImageUrl(slug) + "?thumbnail=true";
```
IEvangelist, line 10:
```suggestion
    public static string GetThumbnailUrl(string slug) => $"{GetImageUrl(slug)}?thumbnail=true";
```

samples/AspireWithAzureFunctions/ImageGallery.FrontEnd/Program.cs
```diff
@@ -43,4 +45,22 @@
 app.MapRazorComponents<App>()
     .AddInteractiveServerRenderMode();
 
+app.MapGet(ImageUrl.RoutePattern, async (string slug, bool? thumbnail, IServiceProvider services, CancellationToken cancellationToken) =>
+{
+    var containerClient = services.GetRequiredKeyedService<BlobContainerClient>(thumbnail == true ? "thumbnails" : "images");
```
IEvangelist, line 50:
```suggestion
    var containerClient = services.GetRequiredKeyedService<BlobContainerClient>(thumbnail ? "thumbnails" : "images");
```

samples/AspireWithAzureFunctions/ImageGallery.FrontEnd/Program.cs
```diff
@@ -43,4 +45,22 @@
...
+    var blobClient = containerClient.GetBlobClient(slug);
+
+    if (!await blobClient.ExistsAsync(cancellationToken))
+    {
+        return Results.NotFound();
+    }
+
+    var properties = (await blobClient.GetPropertiesAsync(cancellationToken: cancellationToken)).Value;
```
IEvangelist, line 58: Dangerous assumption.

samples/AspireWithAzureFunctions/ImageGallery.Functions/ThumbnailGenerator.cs
```diff
@@ -7,15 +7,15 @@
...
+namespace ImageGallery.Functions;
 
 public class ThumbnailGenerator(ILogger<ThumbnailGenerator> logger,
     QueueServiceClient queueServiceClient,
     BlobServiceClient blobServiceClient)
 {
-    private BlobContainerClient containerClient = blobServiceClient.GetBlobContainerClient("thumbnails");
-    private QueueClient resultsQueueClient = queueServiceClient.GetQueueClient("thumbnail-queue");
-    private const int targetHeight = 128;
+    private readonly BlobContainerClient containerClient = blobServiceClient.GetBlobContainerClient("thumbnails");
```
IEvangelist, line 16:
```suggestion
    private readonly BlobContainerClient _containerClient = blobServiceClient.GetBlobContainerClient("thumbnails");
```

samples/AspireWithAzureFunctions/ImageGallery.Functions/ThumbnailGenerator.cs
```diff
@@ -7,15 +7,15 @@
...
+    private readonly QueueClient resultsQueueClient = queueServiceClient.GetQueueClient("thumbnail-queue");
```
IEvangelist, line 17:
```suggestion
    private readonly QueueClient _resultsQueueClient = queueServiceClient.GetQueueClient("thumbnail-queue");
```

samples/AspireWithAzureFunctions/ImageGallery.Functions/ThumbnailGenerator.cs
```diff
@@ -78,7 +78,7 @@ private async Task SendQueueMessageAsync(string name)
         var jsonMessage = JsonSerializer.Serialize(
             new UploadResult(name, true), SerializationContext.Default.UploadResult);
 
-        logger.LogInformation("Signaling upload of {Name}", name);
+        logger.LogDebug("Signaling upload of {Name}", name);
 
         await resultsQueueClient.SendMessageAsync(jsonMessage);
```
IEvangelist, line 83:
```suggestion
        await _resultsQueueClient.SendMessageAsync(jsonMessage);
```

samples/AspireWithAzureFunctions/ImageGallery.FrontEnd/Program.cs
davidfowl: whats that?

samples/AspireWithAzureFunctions/ImageGallery.Functions/ThumbnailGenerator.cs
DamianEdwards: I was keeping it with the existing naming convention this sample uses. Are you saying we should change everything to use `_` prefixing for members?

samples/AspireWithAzureFunctions/ImageGallery.FrontEnd/Program.cs
DamianEdwards: Can't do that because it's nullable.

samples/AspireWithAzureFunctions/ImageGallery.FrontEnd/Program.cs
DamianEdwards: Yeah not sure what you're referring to here.

samples/AspireWithAzureFunctions/ImageGallery.FrontEnd/ImageUrl.cs
DamianEdwards: Personally in this case I find the concat easier to read given it's just two strings and using interpolation results in a sequence like `)}?` in the middle of the string. But I'm not super attached.

samples/AspireWithAzureFunctions/ImageGallery.Functions/ThumbnailGenerator.cs
DamianEdwards: Hmm looking more through the full sample it seems it's not currently consistent. I'll change it all to use underscore prefixing like you suggest.

samples/AspireWithAzureFunctions/ImageGallery.AppHost/Program.cs
```diff
@@ -1,17 +1,31 @@
...
+var storage = builder.AddAzureStorage("storage").RunAsEmulator()
+    .ConfigureConstruct((ResourceModuleConstruct construct) =>
+    {
+        var storageAccount = construct.GetResources().OfType<StorageAccount>().FirstOrDefault(r => r.IdentifierName == "storage")
+            ?? throw new InvalidOperationException($"Could not find configured storage account with name 'storage'");
+        // Storage Account Contributor and Storage Blob Data Owner roles are required by the Azure Functions host
+        construct.Add(storageAccount.CreateRoleAssignment(StorageBuiltInRole.StorageAccountContributor, construct.PrincipalTypeParameter, construct.PrincipalIdParameter));
+        construct.Add(storageAccount.CreateRoleAssignment(StorageBuiltInRole.StorageBlobDataOwner, construct.PrincipalTypeParameter, construct.PrincipalIdParameter));
+        // Ensure that public access to blobs is disabled
+        storageAccount.AllowBlobPublicAccess = false;
```
captainsafia, line 14: The default value is false so we could remove this but I think there's value in the explicitness for this sample.

captainsafia (commented):
LGTM!

samples/AspireWithAzureFunctions/ImageGallery.AppHost/Program.cs
DamianEdwards: Yeah I wasn't actually sure as the doc comments are kind confusing. I interpreted this to mean it defaulted to true?
image

samples/AspireWithAzureFunctions/ImageGallery.FrontEnd/Program.cs
IEvangelist: Wrapping an `await` in parathesis assumes that it returns a non-null value and when you access it with `.Value`, you're saying that you assume for certain that the call to `.GetPropertiesAsync` will happily return. It's cleaner/safer to get the response, check for `null` and then access the `.Value`.

DamianEdwards: That's what we have nullable ref types for though, and `Value` isn't marked as nullable, so this should be fine. There's literally nothing of value on the object returned from `GetPropertiesAsync` other than the `Value` member.

davidfowl: Is the return value nullable @DamianEdwards ?

DamianEdwards: No, and if it were I'd get a compiler message complaining about how I'm using `properties` afterwards. The doc comment for `Value` says it will throw upon access only if the entire response is declared as a nullable response, which it isn't here.

PR #534, opened 2024-11-07
Use string.Create and remove stackalloc

davidfowl:
This is a better way to do the replacement that doesn't assume the length is something that can be allocated on the heap.


dosper7:
> This is a better way to do the replacement that doesn't assume the length is something that can be allocated on the heap.

you mean stack, right? (because of the stackalloc thing, I assume you wanted to say stack instead of heap), otherwise I didn't understood the difference 😅

davidfowl:
Fixed it!

Issue #416, opened 2024-08-15
Create a sample that shows how to use the Keycloak integration


DamianEdwards:
Should show how to configure the following app types to use a Keycloak instance to manage authentication (user, app -> api), and authorization:

- APIs (Minimal APIs and Controller-based)
- Razor Pages
- MVC with Razor Views
- Blazor
  - SSR
  - Server
  - ASP.NET Core hosted WebAssembly

BartoGabriel:
Hi @DamianEdwards

I was reviewing the example with Keycloak, and it was really helpful for the implementation I'm currently working on.

In the example, I noticed that the Blazor SSR application is using the application token (client credentials). While it makes sense that the API only allows authenticated applications to make requests, I think it would be better to use another flow, since user-based authentication is more common in many scenarios.

It would be beneficial for the API to know which user is making the request, what role they have, etc.

Although I have limited experience with Keycloak, I think simplifying the example would help users achieve the following:

Authenticate from the front end.
Have the API handle an authenticated user, including user identification and roles.
Handle token storage and refresh tokens, as access tokens have a short lifespan by default. To simplify the example, I believe using cookies to store tokens would be a good approach.
Regarding the Client Credentials flow, while it's useful, it might be better as a secondary example. For instance, you could show how a console application could call the API using this flow.

As mentioned, I don’t have much experience with Keycloak, but I’d be happy to help in any way I can.

Thanks again for the example—it has been very helpful.

DamianEdwards:
Thanks for the suggestion. I specifically didn't want to have the API require tokens from end users as it has no need for it in this example. The [eShop workshop](https://github.com/dotnet-presentations/eshop-app-workshop/tree/main) does what you're suggesting, as the backend APIs actually deal with user-specific data. I could see this example be expanded to show both flows via different backend APIs, e.g. one that deals with user data and another that doesn't.

Ultimately there's no "one size fits all" solution when dealing with distributed app identity, authentication, and authorization concerns. This sample is really only intended to showcase how the Keycloak integration can be used with Aspire, but not to be an exhaustive reference for Keycloak itself, or ASP.NET Core authentication/authorization in general.

yellow444:
@DamianEdwards Hi!
 I’m developing a test application and used the e-shop template. For authentication, I used your example with Keycloak and noticed an issue with signed certificates: the export command cannot create them in a non-existent folder. I’d like to fix this. I hope my changes will be helpful to others.

#593

PR #593, opened 2024-11-27
Add missing directory creation to fix certificate export issue.

yellow444:
Summary

This PR addresses issue  #416 related to an error that occurs during certificate export when the specified directory does not exist. The issue was caused by the lack of a check or creation step for the directory before attempting to export the certificate.
Root Cause

When exporting the certificate, the CertificateManager throws a System.InvalidOperationException if the target directory is missing. The error message highlights that the directory does not exist and emphasizes the need for careful permission settings when creating it.
Fix

    Added a line to ensure the target directory is created before exporting the certificate:

    Directory.CreateDirectory(tempDir);

Impact

This change prevents the InvalidOperationException and ensures smooth certificate export without requiring manual intervention to create the directory. It also aligns the process with best practices for error handling and directory management.
Testing

    Verified that the directory is created automatically if it does not exist.
    Confirmed successful certificate export without errors.

Additional Notes

This fix improves usability and reliability when working with certificate export functionalities. Feedback and further suggestions are welcome!

### DIFF
--- a/samples/Keycloak/Keycloak.AppHost/HostingExtensions.cs
+++ b/samples/Keycloak/Keycloak.AppHost/HostingExtensions.cs
@@ -86,7 +86,7 @@ public static class HostingExtensions
         {
             Directory.Delete(tempDir, recursive: true);
         }
-
+        Directory.CreateDirectory(tempDir);
         var exportProcess = Process.Start("dotnet", $"dev-certs https --export-path \"{certExportPath}\" --format Pem --no-password");
 
         var exited = exportProcess.WaitForExit(TimeSpan.FromSeconds(5));


yellow444:
#416

DamianEdwards:
Thanks for this. I actually found this too when updating the Node sample to use the dev cert as you can see at: https://github.com/dotnet/aspire-samples/blob/cc0d6b4c28d1f664e74eeccb911c7f23c2950510/samples/Shared/DevCertHostingExtensions.cs#L103-L108

I'll merge this in and when I circle back to the Keycloak sample I'll update it to use the now committed shared logic.

PR #616, opened 2024-12-09
Add explicit System.Runtime.Caching 6.0.1 reference

radical:
.. to fix CG warnings.

### DIFF
--- a/samples/DatabaseContainers/DatabaseContainers.ApiService/DatabaseContainers.ApiService.csproj
+++ b/samples/DatabaseContainers/DatabaseContainers.ApiService/DatabaseContainers.ApiService.csproj
@@ -17,6 +17,9 @@
     <PackageReference Include="Aspire.MySqlConnector" Version="9.0.0" />
     <PackageReference Include="Dapper" Version="2.1.35" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="7.0.0" />
+
+    <!-- explicitly pick newer version with security fix -->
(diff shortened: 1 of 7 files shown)


DamianEdwards:
@radical thanks. Which packages are bringing in the vulnerable version and why haven't they been revved to fix the warning at their level?

joperezr:
> @radical thanks. Which packages are bringing in the vulnerable version and why haven't they been revved to fix the warning at their level?

I also see this CG alert in aspire repo itself, so I'll open a PR to fix that, which relates to your "why haven't they been revved to fix the warning at their level" question. After that goes in, we should check if this PR is still needed.

radical:
> > @radical thanks. Which packages are bringing in the vulnerable version and why haven't they been revved to fix the warning at their level?
> 
> I also see this CG alert in aspire repo itself, so I'll open a PR to fix that, which relates to your "why haven't they been revved to fix the warning at their level" question. After that goes in, we should check if this PR is still needed.

That would be better indeed. The alerts here are because of the Aspire packages - `Aspire.Microsoft.Data.SqlClient`, `Aspire.Microsoft.EntityFrameworkCore.SqlServer` and `Aspire.Hosting.SqlServer`.

joperezr:
Looking closer, seems like System.Runtime.Caching wasn't actually vulnerable and the s[ecurity advisory](https://github.com/dotnet/announcements/issues/327) that had originally marked it as vulnerable has been corrected. Due to this, I have dismissed the alerts and we should be able to close this PR.

PR #667, opened 2025-01-28
Replace use of Path.GetTempDir with Directory.CreateTempSubdirectory

### LINKED ISSUE #666: Replace uses of Path.GetTempDir with Directory.CreateTempSubdirectory instead
Context: https://github.com/dotnet/aspire/issues/7278#issuecomment-2617373361

DamianEdwards:
Fixes #666

Made the directory name contain a sanitized version of the app host application name too.

### DIFF
--- a/samples/Shared/DevCertHostingExtensions.cs
+++ b/samples/Shared/DevCertHostingExtensions.cs
@@ -187,8 +187,9 @@ public static class DevCertHostingExtensions
     {
         // Create a temp directory with a name that is unique to the application and does not change across restarts.
 
+        var appName = Sanitize(builder.Environment.ApplicationName).ToLowerInvariant();
         var appNameHash = builder.Configuration["AppHost:Sha256"]![..10].ToLowerInvariant();
-        var dirName = $"aspire.{appNameHash}";
+        var dirName = $"aspire.{appName}.{appNameHash}";
 
         // Determine the parent directory to create the temp directory in.
         var tempDir = Directory.CreateTempSubdirectory(dirName);
@@ -204,5 +205,22 @@ public static class DevCertHostingExtensions
         }
 
         return fullDirPath;
+    }
+
+    private static char[] _invalidPathChars = Path.GetInvalidPathChars();
+
+    private static string Sanitize(string name)
+    {
+        return string.Create(name.Length, name, static (s, name) =>
+        {
+            var nameSpan = name.AsSpan();
+
+            for (var i = 0; i < nameSpan.Length; i++)
+            {
+                var c = nameSpan[i];
+
+                s[i] = Array.IndexOf(_invalidPathChars, c) == -1 ? c : '_';
+            }
+        });
     }
 }
--- a/samples/Shared/DevCertHostingExtensions.cs
+++ b/samples/Shared/DevCertHostingExtensions.cs
@@ -76,8 +76,8 @@ public static class DevCertHostingExtensions
         // Exports the ASP.NET Core HTTPS development certificate & private key to PEM files using 'dotnet dev-certs https' to a temporary
         // directory and returns the path.
         // TODO: Check if we're running on a platform that already has the cert and key exported to a file (e.g. macOS) and just use those instead.
-        var appNameHash = builder.Configuration["AppHost:Sha256"]![..10];
-        var tempDir = Path.Combine(Path.GetTempPath(), $"aspire.{appNameHash}");
+        var appNameHash = builder.Configuration["AppHost:Sha256"]![..10].ToLowerInvariant();
+        var tempDir = GetOrCreateAppHostTempSubdirectory(builder);
         var certExportPath = Path.Combine(tempDir, "dev-cert.pem");
         var certKeyExportPath = Path.Combine(tempDir, "dev-cert.key");
 
@@ -181,5 +181,28 @@ public static class DevCertHostingExtensions
                 callback(new string(buffer, 0, charsRead));
             }
         }
+    }
+
+    private static string GetOrCreateAppHostTempSubdirectory(IDistributedApplicationBuilder builder)
+    {
+        // Create a temp directory with a name that is unique to the application and does not change across restarts.
+
+        var appNameHash = builder.Configuration["AppHost:Sha256"]![..10].ToLowerInvariant();
+        var dirName = $"aspire.{appNameHash}";
+
+        // Determine the parent directory to create the temp directory in.
+        var tempDir = Directory.CreateTempSubdirectory(dirName);
+        var parentDir = tempDir.Parent ?? throw new InvalidOperationException("User temp directory could not be determined.");
+
+        // Check if the final directory already exists.
+        var fullDirPath = Path.Combine(parentDir.FullName, dirName);
+
+        if (!Directory.Exists(fullDirPath))
+        {
+            // Rename the temp directory to the final directory name so that it's stable across restarts.
+            Directory.Move(tempDir.FullName, fullDirPath);
+        }
+
+        return fullDirPath;
     }
 }


samples/Shared/DevCertHostingExtensions.cs
```diff
@@ -182,4 +182,45 @@ static async Task ConsumeOutput(TextReader reader, Action<string> callback)
...
+        var tempDir = Directory.CreateTempSubdirectory(dirName);
+        var parentDir = tempDir.Parent ?? throw new InvalidOperationException("User temp directory could not be determined.");
+
+        // Check if the final directory already exists.
+        var fullDirPath = Path.Combine(parentDir.FullName, dirName);
+
+        if (!Directory.Exists(fullDirPath))
+        {
+            // Rename the temp directory to the final directory name so that it's stable across restarts.
+            Directory.Move(tempDir.FullName, fullDirPath);
```
eerhardt, line 204: I'm not sure this is fully correct. What this is doing:

1. Create a new directory under TEMP that only the current user has permission to.
2. Move it to a deterministic place directly under TEMP given the app name.

Issues I can think of:

1. If 2 users are trying to run this sample app on the same machine, the first one will create the folder with only permissions to themselves. When the 2nd user runs, the folder will already be there but the user won't have permission to it. So they will get an error.
2. If a malicious user gets on the system first, and adds the folder first (as globally read/write), they could write files to the folder and a different user would just use them.

In general, on Linux it isn't good to use the temp directory for stable files with deterministic names. See https://systemd.io/TEMPORARY_DIRECTORIES/

DamianEdwards: OK seems we can't use temp for this kind of thing then. We'll need to use the user profile or a project artifacts path (bin, obj) instead.

samples/Shared/DevCertHostingExtensions.cs
```diff
@@ -76,8 +76,8 @@ public static IResourceBuilder<TResource> RunWithHttpsDevCertificate<TResource>(
         // Exports the ASP.NET Core HTTPS development certificate & private key to PEM files using 'dotnet dev-certs https' to a temporary
         // directory and returns the path.
         // TODO: Check if we're running on a platform that already has the cert and key exported to a file (e.g. macOS) and just use those instead.
-        var appNameHash = builder.Configuration["AppHost:Sha256"]![..10];
-        var tempDir = Path.Combine(Path.GetTempPath(), $"aspire.{appNameHash}");
+        var appNameHash = builder.Configuration["AppHost:Sha256"]![..10].ToLowerInvariant();
```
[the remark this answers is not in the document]
DamianEdwards: Victim of refactoring

samples/Shared/DevCertHostingExtensions.cs
DamianEdwards: I changed this to use a path under `/obj` instead.

samples/Shared/DevCertHostingExtensions.cs
```diff
@@ -182,4 +180,56 @@ static async Task ConsumeOutput(TextReader reader, Action<string> callback)
...
+        return $"aspire.{appName}.{appNameHash}";
+    }
+
+    private static readonly char[] _invalidPathChars = Path.GetInvalidPathChars();
+
+    private static string Sanitize(string name)
+    {
+        return string.Create(name.Length, name, static (s, name) =>
+        {
+            var nameSpan = name.AsSpan();
```
danmoseley, line 225: what's the advantage of using a span here?

samples/Shared/DevCertHostingExtensions.cs
```diff
@@ -182,4 +180,56 @@ static async Task ConsumeOutput(TextReader reader, Action<string> callback)
...
+        assemblyMetadata?.FirstOrDefault(a => string.Equals(a.Key, key, StringComparison.OrdinalIgnoreCase))?.Value;
+
+    private static string GetAppHostSpecificTempDirPrefix(IDistributedApplicationBuilder builder)
+    {
+        var appName = Sanitize(builder.Environment.ApplicationName).ToLowerInvariant();
+        var appNameHash = builder.Configuration["AppHost:Sha256"]![..10].ToLowerInvariant();
+        return $"aspire.{appName}.{appNameHash}";
+    }
+
+    private static readonly char[] _invalidPathChars = Path.GetInvalidPathChars();
```
danmoseley, line 219: this returns a static readonly char[] already and it will be inlined. there is no reason to put it in a field.

samples/Shared/DevCertHostingExtensions.cs
```diff
@@ -182,4 +180,56 @@ static async Task ConsumeOutput(TextReader reader, Action<string> callback)
...
+
+    private static string Sanitize(string name)
```
danmoseley, line 221: assuming this is not hot and you're already using Linq here how about making this a one liner
```c#
return new string(name.Select(c => Path.GetInvalidPathChars().Contains(c) ? '_' : c).ToArray());
```

samples/Shared/DevCertHostingExtensions.cs
```diff
@@ -182,4 +180,56 @@ static async Task ConsumeOutput(TextReader reader, Action<string> callback)
...
+    {
+        return string.Create(name.Length, name, static (s, name) =>
+        {
+            var nameSpan = name.AsSpan();
+
+            for (var i = 0; i < nameSpan.Length; i++)
+            {
+                var c = nameSpan[i];
+
+                s[i] = Array.IndexOf(_invalidPathChars, c) == -1 ? c : '_';
```
danmoseley, line 231: can be Contains instead of IndexOf.

samples/Shared/DevCertHostingExtensions.cs
```diff
@@ -182,4 +180,56 @@ static async Task ConsumeOutput(TextReader reader, Action<string> callback)
...
+        // macOS: Path.Combine(
+        //          Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".aspnet", "dev-certs", "https"),
+        //          "aspnetcore-localhost-{certificate.Thumbprint}.pfx");
+        // linux: CurrentUser/Root store
+
+        // Create a directory in the project's /obj dir or TMP to store the exported certificate and key
+        var assemblyMetadata = builder.AppHostAssembly?.GetCustomAttributes<AssemblyMetadataAttribute>();
+        var projectDir = GetMetadataValue(assemblyMetadata, "AppHostProjectPath");
+        var objDir = GetMetadataValue(assemblyMetadata, "AppHostProjectBaseIntermediateOutputPath");
+        var dirPath = projectDir is not null && objDir is not null
```
danmoseley, line 196: is AppHostProjectPath expected to be a full path? if so Path.Join may be more appropriate.

samples/Shared/DevCertHostingExtensions.cs
```diff
@@ -182,4 +180,56 @@ static async Task ConsumeOutput(TextReader reader, Action<string> callback)
...
+            ? Path.Combine(projectDir, objDir, "aspire")
+            : Directory.CreateTempSubdirectory(GetAppHostSpecificTempDirPrefix(builder)).FullName;
+
+        if (!Directory.Exists(dirPath))
```
danmoseley, line 200: this is redundant, it's cheaper to just call CreateDirectory

samples/Shared/DevCertHostingExtensions.cs
DamianEdwards: None, will remove.

samples/Shared/DevCertHostingExtensions.cs
danmoseley: You're quite right, sorry (obviously too since it's a public array anyone could mutate)

samples/Shared/DevCertHostingExtensions.cs
DamianEdwards: It is, yes.

samples/Shared/DevCertHostingExtensions.cs
DamianEdwards: It coulde be LINQ, but this is derived from the existing method we have in Aspire.Hosting for volume name sanitization, so I'm inclined to just keep it this way for now.

samples/Directory.Build.targets
```diff
@@ -0,0 +1,10 @@
+<Project>
```
eerhardt, line 1: It wouldn't hurt to stick a comment in here in case someone stumbles in and asks what this is for.

samples/Shared/DevCertHostingExtensions.cs
```diff
@@ -17,96 +18,162 @@ public static class DevCertHostingExtensions
     /// Use <see cref="ResourceBuilderExtensions.WithHttpsEndpoint{TResource}"/> to configure an HTTPS endpoint.
     /// </remarks>
     public static IResourceBuilder<TResource> RunWithHttpsDevCertificate<TResource>(
-        this IResourceBuilder<TResource> builder, string certFileEnv, string certKeyFileEnv, Action<string, string>? onSuccessfulExport = null)
+        this IResourceBuilder<TResource> builder, CertificateFileFormat certificateFileFormat, string certFileEnv, string? certPasswordOrKeyEnv, Func<IServiceProvider, string, string?, Task>? onSuccessfulExport = null)
```
eerhardt, line 21: The name `certPasswordOrKeyEnv` is a bit confusing. is there a better API or documentation or something we can do here?

samples/Shared/DevCertHostingExtensions.cs
```diff
@@ -17,96 +18,162 @@ public static class DevCertHostingExtensions
...
+            if (File.Exists(certExportPath) && File.Exists(certKeyExportPath))
+            {
+                // Certificate already exported, return the path.
+                logger.LogDebug("Using previously exported dev cert PEM file '{CertPath}' and key file '{CertKeyPath}'", certExportPath, certKeyExportPath);
+                return (true, certExportPath, certKeyExportPath);
+            }
+
+            if (File.Exists(certExportPath))
+            {
+                logger.LogTrace("Previously exported key file is present but dev cert file '{CertPath}' is missing. Deleting dev cert file and re-exporting.", certExportPath);
```
eerhardt, line 159: Why do we warn above in PfxWithPassword, but only trace here?

eerhardt (approved):

PR #724, opened 2025-03-07
Update all OTEL packages

danmoseley:
Update all OTEL packages. Two of them have a CVE; rather than reasoning, I've updated everything to latest.


DamianEdwards:
So far we've chosen to keep the samples inline with https://github.com/dotnet/aspire/issues/7123

Updating all the OTel packages means all the projects here (vast majority of which are `net8.0`) have all the `Microsoft.Extensions.*` deps lifted out of .NET 8 (LTS).

danmoseley:
ready for review again.

DamianEdwards:
@danmoseley looks like versions of Otel libs you're now referencing need to be mirrored or something? There's nuget restore warnings about unmatched versions and then it resolves one with a vulnerability.

danmoseley:
Yup (https://dev.azure.com/dnceng/internal/_build?definitionId=931&_a=summary) was waiting for it. I"ll rerun now it's done

DamianEdwards:
Actually this repo uses nuget.org as it's samples, not product code. I think you've just got a bad version in there for the `OpenTelemetry.Instrumentation.GrpcNetClient` package

PR #731, opened 2025-03-09
Onboard to the GitHub workflow based issue-labeler

jeffhandley:
We have a new implementation of [dotnet/issue-labeler](https://github.com/dotnet/issue-labeler) available that is implemented entirely using GitHub workflows. This approach allows for self-service onboarding and re-training of the prediction models. [Onboarding](https://github.com/dotnet/issue-labeler/wiki/Onboarding#) instructions are documented and anyone with write permission to this repository should be able to work through the process of the initial model training, but @RussKie and I are both familiar with the process.

I configured the "Labeler: Train Models" workflow to allow input of the repository to train from in case we want to train this repository from the content in dotnet/aspire.

### DIFF
--- a/.github/workflows/labeler-build-predictor.yml
+++ b/.github/workflows/labeler-build-predictor.yml
@@ -1,0 +1,17 @@
+name: "Labeler: Build Predictor App"
+
+on:
+  # Allow dispatching the workflow via the Actions UI
+  workflow_dispatch:
+    inputs:
+      rebuild:
+        description: "Force a rebuild of the app"
+        type: boolean
+
+jobs:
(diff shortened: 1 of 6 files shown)


danmoseley (approved):
ha, I was about to onboard this repo

danmoseley:
I kicked off training to "staging" -- hopefully that's correct.

wrt the repo to train on -- this one primarily, but I could imagine augmenting with dotnet/aspire. Just not only dotnet/aspire. Can it do that? It's not important, if it can't.

With the old labeler, we did this in one case (mixed in two repos to train) -- I forget which, though.

PR #736, opened 2025-03-11
Add full stack Node.js solution

IEvangelist:
Add full stack Node.js solution, this PR:

- Demonstrates an end-to-end JavaScript solution, that's orchestrated by .NET Aspire.
- Adds a simple app host project.
- Adds an `api` express-based Node.js app that serves random weather forecast data (much like the ASP.NET Core Minimal API template).
- Adds an `app` that's copied from the existing React bits in the JavaScript samples.

Contributes to https://github.com/dotnet/docs-aspire/issues/2767


samples/FullStackJS/app/package.json
```diff
@@ -0,0 +1,31 @@
+{
```
davidfowl, line 1: Why not vite?

IEvangelist: Because the frontend was based on an existing frontend in the samples, that's using webpack. We have other samples with vite already as well.

samples/FullStackJS/FullStackJS.AppHost/FullStackJS.AppHost.csproj
```diff
@@ -13,7 +13,9 @@
 
   <ItemGroup>
     <PackageReference Include="Aspire.Hosting.AppHost" Version="9.1.0" />
+    <PackageReference Include="Aspire.Hosting.Azure" Version="9.1.0" />
```
DamianEdwards, line 16: Why?

samples/FullStackJS/FullStackJS.AppHost/Program.cs
```diff
@@ -0,0 +1,23 @@
+using Microsoft.Extensions.Hosting;
+
+var builder = DistributedApplication.CreateBuilder(args);
+
+var api = builder.AddNpmApp("api", "../api")
+                 .WithNpmPackageInstallation()
+                 .WithExternalHttpEndpoints()
+                 .PublishAsDockerFile();
+
+_ = builder.Environment.IsDevelopment()
```
davidfowl, line 10:
```suggestion
_ = builder.ExecutionContext.IsPublishMode
```

samples/FullStackJS/app/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM node:23-alpine as build
```
davidfowl, line 1: We should look into optimizations we can do with node docker files. They are many.

samples/FullStackJS/FullStackJS.AppHost/FullStackJS.AppHost.csproj
IEvangelist: We were testing ACA provisioning changes, we found a bug in `azd`. This can be removed now.

PR #792, opened 2025-05-13
Update to 9.3

DamianEdwards:


samples/AspireShop/AspireShop.ServiceDefaults/AspireShop.ServiceDefaults.csproj
```diff
@@ -11,8 +11,8 @@
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
 
-    <PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="9.3.0" />
-    <PackageReference Include="Microsoft.Extensions.ServiceDiscovery" Version="9.2.0" />
+    <PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="9.4.0" />
+    <PackageReference Include="Microsoft.Extensions.ServiceDiscovery" Version="9.3.0" />
```
joperezr, line 15: NIT: I know not technically part of your changes but there is a 9.5.0 version now of Microsoft.Extensions.Http Resilience that shipped last week.

joperezr (commented):
Should our templates be called AppHost.cs going forward as opposed to Program.cs?

joperezr (approved):
Minor comment but other than that things LGTM.

DamianEdwards:
>Should our templates be called AppHost.cs going forward as opposed to Program.cs?

Yeah we made that change in 9.3 😄

samples/AspireShop/AspireShop.ServiceDefaults/AspireShop.ServiceDefaults.csproj
DamianEdwards: thanks, I'll pick that up in the next dependency update

IEvangelist:
> > Should our templates be called AppHost.cs going forward as opposed to Program.cs?
> 
> Yeah we made that change in 9.3 😄

Wait...why did this change? The project name is AppHost, and this was originally the `Program`. I don't like that change. It's not really something we do anywhere else and hides that fact that it's the entry point of the executable project.

joperezr:
> > > Should our templates be called AppHost.cs going forward as opposed to Program.cs?
> 
> > 
> 
> > Yeah we made that change in 9.3 😄
> 
> 
> 
> Wait...why did this change? The project name is AppHost, and this was originally the `Program`. I don't like that change. It's not really something we do anywhere else and hides that fact that it's the entry point of the executable project.
> 
> 
> 
> [This also broke docs build](https://buildapi.docs.microsoft.com/Output/PullRequest/88fc8115-034d-82dd-b8c7-be05f0097b39/202505211205198088-3514/BuildReport?accessString=c9226bc71db41dee92ce637e2afae515ea849e28c15465907af7e877f8e1afd5), I didn't know this changed.
> 
> 
> 
> It feels like now we're have to disambiguate between the "app host" concept, the project file and the class. I really think this adds no value, but causes a point of confusion and ambiguity.

Happy to discuss this. While I'm not certain, I suppose the reasoning was that, before this, you'd have 3 different Program.cs at least in a starter app, and when developing and jumping in between files it does happen often that you open the wrong file (if you are clicking on the opened tabs as opposed to solution explorer). Having the name of the file explicitly say that this is the apphost file disambiguates this. Again, not sure if this is why it was done, but I've personally hit this issue multiple times so it is nice to now have this fixed.

PR #1008, opened 2025-08-22
Add .NET 8.0 runtime support to devcontainer configuration

sliekens:
The samples still use `TargetFramework: net8.0` so it is necessary to have a .NET 8.0 runtime installed to run the samples. (The .NET 9.0 SDK only contains the .NET 9.0 runtime.)

The "version" field is set to "none" to avoid overwriting the SDK version from the base image.
The "aspNetCoreRuntimeVersions" matches the TFM used by the samples.


danmoseley (approved):
thanks

PR #1116, opened 2025-09-23
Update to 9.5

DamianEdwards:
Updates samples to 9.5.

Needs changes to `nuget.config` reverted before merging.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -3,21 +3,8 @@
   <packageSources>
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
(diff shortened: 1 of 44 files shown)


NuGet.config
```diff
@@ -3,8 +3,21 @@
...
+      <package pattern="Microsoft.Extensions.ServiceDiscovery*" />
+    </packageSource>
+    <packageSource key="nuget.org">
+      <package pattern="*" />
+    </packageSource>
+  </packageSourceMapping>
+  <config>
+    <add key="globalPackagesFolder" value=".nugetpackages" />
+  </config>
 </configuration>
```
davidfowl, line 23:
```suggestion
  <packageSources>
    <clear />
    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
  </packageSources>
  <solution>
    <add key="disableSourceControlIntegration" value="true" />
  </solution>
</configuration>
```

PR #1217, opened 2025-10-21
Rename WPF+WinForms AppHost back to Program

jfversluis:
Probably during a bulk rename of Program.cs to AppHost.cs the Program.cs files of the WPF and WinForms samples also got renamed.

This PR changes those back


DamianEdwards:
Thanks, but not required anymore as this got resolved as part of the updates for 13.0

PR #1289, opened 2025-11-25
Disable parallel build for now

DamianEdwards:
Disable parallel build while we investigate the NuGet SDK resolver issue.


eerhardt (approved):
I hope we can figure this out quick!

PR #1302, opened 2025-11-29
Example project for handling local OS detection and specific wireup

dylan-asos:
Engineers may have a mix of Windows and Mac machines as their local development environment. It should be pull and run, regardless of the target architecture. 

Certain Azure services require specific setup when running emulated under ARM, notably around Cosmos / SQL. Other services play nicely on both, e.g azurite for storage. 

This example attempts to demonstrate how you might go about detecting the local environment, and calling into the appropriate setups where there are differences. The demo includes a minimal API that has the following dependencies

- Azure Service Bus
- Cosmos DB
- Storage Blobs + Storage Queue
- SQL Server

The API exposes a single endpoint that will attempt operations on all the dependencies and return some data it reads from Cosmos.

### DIFF
--- a/samples/cross-platform-devx/CrossPlatform.Web.Api/Data/CqrsModule.cs
+++ b/samples/cross-platform-devx/CrossPlatform.Web.Api/Data/CqrsModule.cs
@@ -12,7 +12,7 @@ public static class CqrsModule
         services.AddTransient<IQueryService, QueryService>();
         services.AddTransient<IRequestHandlerResolver>(ctx => new RequestHandlerResolver(ctx.GetRequiredService));
         
-        // API specific queries 
+        // This project specific queries 
         services.AddTransient<GetEnrichedGeographiesQueryHandler>();
         services.AddTransient<QueryHandler<GetEnrichedGeographiesQuery, IEnumerable<Geography>>>(sp => 
             sp.GetRequiredService<GetEnrichedGeographiesQueryHandler>());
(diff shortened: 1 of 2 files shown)


dylan-asos:
@dotnet-policy-service agree

dylan-asos:
@DamianEdwards what do you reckon, worth sharing wider? I found some services particularly gnarly on ARM, wasn't totally clear what combination of config would put me into a healthy state. 

Cosmos/ASB/SQL seem to need a little extra TLC - any others that differ between Mac + Win? Appreciate it's fairly fast moving, but would a project that basically allows you to demo "here are all the spicy components that need that little extra" be useful?

davidfowl:
It's not clear to me that this is the right approach, we may just need to get the teams that own these containers to fix them so that they run everywhere. Most of these services were born on windows this is the side effect.

Does the SQL 2025 image work on MacOS now? That would be a better alternative as SQL edge is retired:
https://learn.microsoft.com/en-us/lifecycle/products/azure-sql-edge

Notice that azurite work well everywhere 😄

dylan-asos:
> Does the SQL 2025 image work on MacOS now?

I couldn't get it going, both 2022 and 2025 giving me :

```
 <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
 <jemalloc>: (This is the expected behaviour if you are running under QEMU)
 /opt/mssql/bin/sqlservr: Invalid mapping of address 0x400534d000 in reserved address space below 0x400000000000. Possible causes:
```

with 2022
image

with 2025
image

Appreciate edge is retired but was the only image I could get both ASB and SQL running against, with all reporting healthy + the API able to execute a request.

Absolutely agree you'd want a single API across all - if teams can fix them, perfect. Until then, do we need an if-mac-else-windows branch for these particular services?

I can close PR if don't want in samples + happy to leave the fork as a reference

dylan-asos:
Closing PR - assuming we'll end up with a single API and no need for the OS specific bootstrapping. Will leave fork in place, might help anyone until then

PR #1573, opened 2026-04-27
Add policies to label and auto-merge Dependabot PRs

IEvangelist:
Adds two GitOps policy configuration files under `.github/policies/` so that Dependabot pull requests are automatically labeled and auto-merged once CI is green.

Modeled after the equivalent policies in [dotnet/docs](https://github.com/dotnet/docs/tree/main/.github/policies).

### Files

- **`.github/policies/label-prs.yml`** — On open, labels Dependabot-authored PRs with `dependencies` and `auto-merge`, and Dependabot-authored issues with `dependencies`.
- **`.github/policies/auto-merge.yml`** — When a PR targeting `main` is labeled `auto-merge`, the bot approves it and enables GitHub's native auto-merge (squash). GitHub only completes the merge once all required status checks pass, so only green PRs land. Removing the label disables auto-merge.

### Why two files

This mirrors the structure used in dotnet/docs and keeps responsibilities separated:

- **Labeling** is the policy that decides *which* PRs are eligible for auto-merge (only Dependabot's, by default).
- **Auto-merge** is purely label-driven, so a maintainer can also opt a non-Dependabot PR in by adding the `auto-merge` label, or out by removing it.

### Behavior

- Only PRs targeting `main` are auto-merged.
- `triggerOnOwnActions: true` is set on the auto-merge rule so that the policy bot's own `labelAdded` event (from `label-prs.yml`) triggers the merge configuration.
- GitHub's branch protection / required checks remain the gate that determines "green"; this policy never bypasses CI.

### DIFF
--- a/.github/policies/auto-merge.yml
+++ b/.github/policies/auto-merge.yml
@@ -1,0 +1,36 @@
+id: auto.merge
+name: GitOps.PullRequestIssueManagement
+description: Auto-merge pull requests labeled 'auto-merge' once CI is green
+owner:
+resource: repository
+disabled: false
+where:
+configuration:
+  resourceManagementConfiguration:
+    eventResponderTasks:
+
+      - description: >-
+          Approve and auto-squash-merge PRs to main labeled 'auto-merge'.
+          The label is only applied automatically to Dependabot PRs (see label-prs.yml),
+          so this effectively gates auto-merge to bot-authored or maintainer-opted-in PRs.
+          GitHub will only complete the merge once all required status checks pass.
+        triggerOnOwnActions: true
+        if:
+          - payloadType: Pull_Request
+          - labelAdded:
+              label: auto-merge
+          - targetsBranch:
+              branch: main
+        then:
+          - enableAutoMerge:
+              mergeMethod: Squash
+          - approvePullRequest:
+              comment: "Approved; this PR will merge when all required status checks pass."
+
+      - description: Disable auto-merge when the 'auto-merge' label is removed
+        if:
+          - payloadType: Pull_Request
+          - labelRemoved:
+              label: auto-merge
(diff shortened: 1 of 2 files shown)


IEvangelist:
Thanks for the review, but most of these suggestions don't apply to the [GitOps policy service](https://eng.ms/docs/products/gitops) that consumes these files — they're written against generic YAML / GitHub-API assumptions rather than the policy schema. The two files already match the canonical pattern used by [`dotnet/docs/.github/policies`](https://github.com/dotnet/docs/tree/main/.github/policies), so I'm not making changes. Quick rationale per comment:

- **`owner:` / `where:` are null** — These are standard placeholder keys in the GitOps repository-policy template. The canonical `dotnet/docs` policies emit them exactly the same way (empty), and the policy service expects this shape. Removing them would diverge from the template.
- **Add a re-approval responder on PR sync (stale approvals)** — Not part of the canonical pattern. `enableAutoMerge` configures GitHub's native auto-merge, which re-evaluates required checks after each push; the canonical `auto-merge.yml` in `dotnet/docs` deliberately fires only on `labelAdded`.
- **Label prerequisite (`dependencies`, `auto-merge` must pre-exist)** — The policy service's `addLabel` action creates the label if it's missing; the canonical Dependabot-labeling example in `dotnet/docs` documents no such prerequisite.
- **Scope `disableAutoMerge` to `main`** — The canonical `dotnet/docs/auto-merge.yml` intentionally leaves the disable rule unscoped. It's idempotent and acts as a defense-in-depth "if the label is removed, do not auto-merge" regardless of target branch.

Happy to revisit any of these if there's a concrete eng.ms GitOps doc that contradicts the above.

eerhardt (approved):

PR #1604, opened 2026-05-15
Harden the Go API sample

DamianEdwards:
## Summary
- Add JSON request body size limiting, item name/count validation, bounded HTTP server timeouts, and loopback-only local binding for `samples/golang-api`.
- Pin the Go API runtime image to Alpine 3.21, run the container as a non-root user, and set the container bind host to `0.0.0.0` so publish/container ingress still works.
- Document sample security limitations and link relevant Go, Docker, and OWASP guidance.

## Validation
- `go fmt ./...`, `go test ./...`, and `go build ./...` using portable Go 1.23.12 because local `go`/`gofmt` are not installed on PATH.
- `docker build --quiet --tag golang-api-bind-check samples\golang-api\api`.

## Aspire verification
- Command: `aspire run` from `samples\golang-api` with portable Go 1.23.12 on PATH.
- Result: AppHost started; dashboard available at `https://localhost:52225/login?t=8ab3aa69913b73430e08b7545acda3fb`.
- Resource status: `api` executable running from `go run main.go`; `api-go-mod-installer` completed/finished.
- Aspire resource URL: `http://localhost:55008`.
- Direct Go process listener: `main.exe` listened only on `127.0.0.1:61189`; the Aspire/DCP endpoint listened on localhost (`127.0.0.1`/`::1`) at `55008`.
- API checks: `GET /`, `GET /health`, and `GET /items` all returned HTTP 200; `/health` returned `{ "status": "healthy" }`.
- Edge screenshots saved locally under the verification session artifacts: `C:\Users\dedward\.copilot\workspaces\f81cff41-43a0-465d-8d45-680462a930cf\artifacts\golang-api-loopback-root-edge.png` and `C:\Users\dedward\.copilot\workspaces\f81cff41-43a0-465d-8d45-680462a930cf\artifacts\golang-api-loopback-aspire-resources-edge.png`.

### DIFF
--- a/samples/golang-api/README.md
+++ b/samples/golang-api/README.md
@@ -73,3 +73,19 @@ else
 - `POST /items` - Create new item
 - `PUT /items/{id}` - Update item
 - `DELETE /items/{id}` - Delete item
+
+## Security Notes
+
+This sample keeps the API intentionally small for demo purposes:
+
+- It does not implement authentication or authorization.
+- Data is stored only in memory and is lost when the process restarts.
+- The request body, item name, and in-memory item count limits are illustrative safeguards, not production capacity planning.
+- External HTTP endpoints are enabled to make the demo easy to run and inspect.
+- Production services should add real authentication, authorization, rate limiting, persistent storage, and monitoring appropriate for their threat model.
+
+Related references:
+
+- [Go `net/http.Server` docs](https://pkg.go.dev/net/http#Server)
+- [Dockerfile best practices: user](https://docs.docker.com/build/building/best-practices/#user)
+- [OWASP API Security Top 10](https://owasp.org/API-Security/editions/2023/en/0x11-t10/)
--- a/samples/golang-api/api/Dockerfile
+++ b/samples/golang-api/api/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.23-alpine AS builder
+FROM golang:1.23-alpine3.21 AS builder
 
 WORKDIR /app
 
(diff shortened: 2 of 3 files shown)


IEvangelist (approved):
Thanks Damian! 🙏

No blockers, but a handful of real things worth a follow-up:

1. **`decodeJSONBody` (main.go:143)** — wraps `r.Body` with `http.MaxBytesReader` but no `defer r.Body.Close()`. If the limit is exceeded mid-stream the underlying body may not be drained.
2. **`json.NewEncoder(w).Encode(...)` errors silently ignored** at lines 203, 211, 217, 236, 253, 258, 271. Encoding failures leave clients with incomplete responses and no signal. At minimum log them (headers are already written so we can't redirect to a different status).
3. **No graceful shutdown** — `ListenAndServe` + `log.Fatal` means SIGTERM aborts in-flight requests. Aspire/containers will typically send SIGTERM. Worth a signal handler + `server.Shutdown(ctx)`.
4. **Dockerfile USER applied after COPY** (line 24) — the binary ends up `root:root`. `COPY --from=builder --chown=app:app …` would be tidier.

The hardening itself is solid:
- ✅ HTTP server timeouts (ReadTimeout, ReadHeaderTimeout, WriteTimeout, IdleTimeout) — addresses slowloris
- ✅ `http.MaxBytesReader` body cap
- ✅ Input validation (name length, UTF-8 rune counting)
- ✅ Store capacity limit
- ✅ Non-root container user, pinned Alpine, loopback-only default with `HOST=0.0.0.0` override

Merging — will follow up on the items above.

PR #1689, opened 2026-06-09
Use MigrationsAssembly(Assembly) overload in database-migrations sample

### LINKED ISSUE #1456: A new MigrationsAssembly overload available
A minor improvement (avoiding literal strings😄) with a new "underrated" overload:
```cs
 sqlOptions.MigrationsAssembly(typeof(MigrationService).Assembly) 
```

https://github.com/dotnet/aspire-samples/blob/43d1746a2e05e3b25f4dee1709b1125969e857d9/samples/database-migrations/DatabaseMigrations.MigrationService/Program.cs#L12

IEvangelist:
## Summary

Addresses #1456 by switching the database-migrations sample to the typed `MigrationsAssembly(Assembly)` overload instead of passing a literal assembly-name string.

```csharp
// Before
sqlOptions.MigrationsAssembly("DatabaseMigrations.MigrationService")

// After
sqlOptions.MigrationsAssembly(typeof(ApiDbInitializer).Assembly)
```

The migrations live in the `DatabaseMigrations.MigrationService` assembly, so the assembly is resolved via `typeof(ApiDbInitializer).Assembly` (a public type already in scope from that assembly). This avoids the magic string and keeps the reference refactor-safe.

The `MigrationsAssembly(Assembly)` overload is available in EF Core 9+ ([API docs](https://learn.microsoft.com/dotnet/api/microsoft.entityframeworkcore.infrastructure.relationaldbcontextoptionsbuilder-2.migrationsassembly)), and this sample targets net10.0 / EF Core 10, so it's applicable.

## Testing

- `dotnet build` on `DatabaseMigrations.MigrationService` — succeeds with 0 warnings, 0 errors.

Fixes #1456

### DIFF
--- a/samples/database-migrations/DatabaseMigrations.MigrationService/Program.cs
+++ b/samples/database-migrations/DatabaseMigrations.MigrationService/Program.cs
@@ -9,7 +9,7 @@ builder.AddServiceDefaults();
 
 builder.Services.AddDbContextPool<MyDb1Context>(options =>
     options.UseSqlServer(builder.Configuration.GetConnectionString("db1"), sqlOptions =>
-        sqlOptions.MigrationsAssembly("DatabaseMigrations.MigrationService")
+        sqlOptions.MigrationsAssembly(typeof(ApiDbInitializer).Assembly)
     ));
 builder.EnrichSqlServerDbContext<MyDb1Context>();
 


davidfowl:
Are we using the new EF integration? 

cc @AndriySvyryd

IEvangelist:
> Are we using the new EF integration?

Yes — both projects use the Aspire EF Core SQL Server integration (`Aspire.Microsoft.EntityFrameworkCore.SqlServer`, 13.4.3). The API service wires it up with the `builder.AddSqlServerDbContext<MyDb1Context>("db1")` one-liner. The migration service registers the context itself with `AddDbContextPool(...)` and then calls `builder.EnrichSqlServerDbContext<MyDb1Context>()` — it uses the explicit registration specifically so it can set `MigrationsAssembly` on the SQL Server options.

This PR (#1456) only swaps the literal `"DatabaseMigrations.MigrationService"` string for the typed `MigrationsAssembly(typeof(ApiDbInitializer).Assembly)` overload; it doesn't change how the integration is wired up. If you'd prefer the migration service consolidate onto `AddSqlServerDbContext` with its options callback instead of the explicit `AddDbContextPool` + `EnrichSqlServerDbContext` pattern, I'm happy to do that as a follow-up.

AndriySvyryd:
> Are we using the new EF integration?
> 
> cc @AndriySvyryd

Filed https://github.com/microsoft/aspire-samples/issues/1694

AndriySvyryd (approved):

PR #1691, opened 2026-06-09
Serve the Azure Functions host over HTTPS in the Functions sample

### LINKED ISSUE #752: Ensure HTTPS works in the Azure Functions sample
From [comment](https://github.com/dotnet/aspire-samples/issues/519#issuecomment-2741065119)

IEvangelist:
Fixes #752

## Problem

In the `aspire-with-azure-functions` sample, the local Azure Functions host (`func host start`) was only ever served over **HTTP**. Simply adding `--useHttps` to the Functions launch profile is not enough — the .NET build of Azure Functions Core Tools cannot auto‑generate a certificate and fails to start:

```
Auto cert generation is currently not working on the .NET Core build.
```

`func host start --useHttps` requires an explicit **password‑protected** PFX via `--cert`/`--password`.

## Fix

- **`ImageGallery.Functions/Properties/launchSettings.json`** — add `--useHttps` so Aspire registers an `https` endpoint for the Functions resource.
- **`ImageGallery.AppHost/AppHost.cs`** —
  - Point `WithUrlForEndpoint` at the `https` endpoint.
  - In **run mode only**, export the trusted ASP.NET Core developer certificate via Aspire's `WithHttpsDeveloperCertificate(...)` and hand the resulting PFX to the Functions host with `WithHttpsCertificateConfiguration(...)` (`--cert <pfx> --password <generated>`). The export password is an ephemeral, per‑run secret parameter, so nothing is committed. The wiring is guarded by `IsRunMode` so it never affects publish/deploy.

Effective local launch becomes:
```
func host start --port 7221 --useHttps --cert <exported-dev-cert>.pfx --password <generated>
```

## Verification (local)

Tested locally with .NET 10, Aspire 13.4.3, Azure Functions Core Tools v4.12.0, Docker (Azurite):

**Before** — `curl http://localhost:7221/admin/host/status` → `200`; HTTPS TLS handshake fails (`SEC_E_INVALID_TOKEN`).

**After**
- `func` launches with `--useHttps --cert <pfx> --password <generated>` and listens on `:7221`.
- `curl https://localhost:7221/admin/host/status` (no `-k`) → **`200 OK`** with `{"state":"Running","version":"4.1048..."}` — i.e. served over HTTPS using the **trusted** dev cert.
- Plain `http://localhost:7221/...` now fails (port is HTTPS‑only).
- Frontend serves over HTTPS (`200`).
- End‑to‑end regression check: uploaded a 400×300 image to the `images` container → the `ThumbnailGenerator` blob trigger produced a 170×128 thumbnail in `thumbnails`. The pipeline is unaffected.

## Prerequisite

A trusted ASP.NET Core developer certificate (`dotnet dev-certs https --trust`) — already part of the Aspire prerequisites linked in the sample README.

### DIFF
--- a/samples/aspire-with-azure-functions/ImageGallery.AppHost/AppHost.cs
+++ b/samples/aspire-with-azure-functions/ImageGallery.AppHost/AppHost.cs
@@ -34,7 +34,35 @@ var functions = builder.AddAzureFunctionsProject<Projects.ImageGallery_Functions
                             // Queue Data Contributor role is required to send messages to the queue
                             StorageBuiltInRole.StorageQueueDataContributor)
                        .WithHostStorage(storage)
-                       .WithUrlForEndpoint("http", u => u.DisplayText = "Functions App");
+                       .WithUrlForEndpoint("https", u => u.DisplayText = "Functions App");
+
+if (builder.ExecutionContext.IsRunMode)
+{
+    // The Functions project's launchSettings.json passes '--useHttps' to the local Azure Functions host.
+    // Azure Functions Core Tools can't auto-generate a certificate on the .NET build of the tools, so the
+    // trusted ASP.NET Core developer certificate is exported to a password-protected PFX and handed to
+    // 'func host start' via its --cert/--password options.
+    var functionsCertPassword = builder.AddResource(
+        ParameterResourceBuilderExtensions.CreateGeneratedParameter(
+            builder,
+            "functions-cert-password",
+            secret: true,
+            parameterDefault: new GenerateParameterDefault()));
+
+    // The developer certificate APIs used below to serve the Azure Functions host over HTTPS are experimental.
+#pragma warning disable ASPIRECERTIFICATES001
+    functions
+        .WithHttpsDeveloperCertificate(functionsCertPassword)
+        .WithHttpsCertificateConfiguration(context =>
+        {
+            context.Arguments.Add("--cert");
+            context.Arguments.Add(context.PfxPath);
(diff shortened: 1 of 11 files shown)


samples/aspire-with-azure-functions/ImageGallery.AppHost/AppHost.cs
```diff
@@ -34,7 +37,31 @@
...
+
+if (builder.ExecutionContext.IsRunMode)
+{
+    // The Functions project's launchSettings.json passes '--useHttps' to the local Azure Functions host.
+    // Azure Functions Core Tools can't auto-generate a certificate on the .NET build of the tools, so the
+    // trusted ASP.NET Core developer certificate is exported to a password-protected PFX and handed to
+    // 'func host start' via its --cert/--password options. The PFX must be encrypted with a non-empty
+    // password, so an ephemeral one is generated for local development.
+    var functionsCertPassword = builder.AddParameter(
+        "functions-cert-password", () => Guid.NewGuid().ToString("N"), secret: true);
```
DamianEdwards: Given this is only done in run mode, is there any benefit to this being a parameter? Is it just so we can mark it a secret so it gets hidden in the dashboard UI? Parameters show up in the UI and add code here so just want to make sure there's a reasonable benefit.

nit: wrapping

sebastienros: Also why use a guid when parameters already handle password generation

PR #1891, opened 2026-08-14
Add MailDev and MailKit custom integration sample

IEvangelist:
## Summary

- add an Aspire 13.5 custom MailDev hosting integration exported through ATS
- add a MailKit client integration with health checks, tracing, metrics, and secure credentials
- provide a runnable TypeScript AppHost plus a compile-validated C# equivalent
- add a newsletter API and focused hosting/client tests

## Validation

- MailDev hosting tests: 4 passed
- MailKit client tests: 8 passed
- TypeScript AppHost compile and lint passed
- end-to-end subscribe/unsubscribe flow verified against MailDev
- vulnerable package checks found no known vulnerabilities

### DIFF
--- a/samples/maildev-mailkit/MailDev.Hosting/MailDev.Hosting.csproj
+++ b/samples/maildev-mailkit/MailDev.Hosting/MailDev.Hosting.csproj
@@ -1,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <EnableAspireIntegrationAnalyzers>true</EnableAspireIntegrationAnalyzers>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Aspire.Hosting" Version="13.5.0" />
+  </ItemGroup>
+
+</Project>
--- a/samples/maildev-mailkit/MailDev.Hosting/MailDev.Hosting.csproj
+++ b/samples/maildev-mailkit/MailDev.Hosting/MailDev.Hosting.csproj
@@ -6,10 +6,12 @@
     <Nullable>enable</Nullable>
     <EnableAspireIntegrationAnalyzers>true</EnableAspireIntegrationAnalyzers>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <!-- The Aspire Type System export attribute is an evaluation feature in 13.4.x; opt in explicitly. -->
+    <NoWarn>$(NoWarn);ASPIREATS001</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Aspire.Hosting" Version="13.5.0" />
+    <PackageReference Include="Aspire.Hosting" Version="13.4.0" />
   </ItemGroup>
 
 </Project>
--- a/samples/maildev-mailkit/MailDev.Hosting/MailDev.Hosting.csproj
+++ b/samples/maildev-mailkit/MailDev.Hosting/MailDev.Hosting.csproj
@@ -5,6 +5,7 @@
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <EnableAspireIntegrationAnalyzers>true</EnableAspireIntegrationAnalyzers>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
   <ItemGroup>
--- a/samples/maildev-mailkit/MailDev.Hosting/MailDevResource.cs
+++ b/samples/maildev-mailkit/MailDev.Hosting/MailDevResource.cs
@@ -1,0 +1,36 @@
+namespace Aspire.Hosting.ApplicationModel;
+
+/// <summary>Represents a MailDev container resource.</summary>
+[AspireExport]
+public sealed class MailDevResource(
+    [ResourceName] string name,
+    ParameterResource? username,
+    ParameterResource password)
+    : ContainerResource(name), IResourceWithConnectionString
+{
+    internal const string HttpEndpointName = "http";
+    internal const string SmtpEndpointName = "smtp";
+
+    private const string DefaultUsername = "mail-dev";
+    private EndpointReference? _smtpEndpoint;
+
+    /// <summary>Gets the optional MailDev SMTP username parameter.</summary>
+    public ParameterResource? UsernameParameter { get; } = username;
+
+    /// <summary>Gets the MailDev SMTP password parameter.</summary>
+    public ParameterResource PasswordParameter { get; } = password;
+
+    internal ReferenceExpression UsernameReference =>
+        UsernameParameter is not null
+            ? ReferenceExpression.Create($"{UsernameParameter}")
+            : ReferenceExpression.Create($"{DefaultUsername}");
+
+    /// <summary>Gets the MailDev SMTP endpoint.</summary>
+    public EndpointReference SmtpEndpoint =>
+        _smtpEndpoint ??= new(this, SmtpEndpointName);
+
+    /// <inheritdoc />
+    public ReferenceExpression ConnectionStringExpression =>
+        ReferenceExpression.Create(
+            $"Endpoint=smtp://{SmtpEndpoint.Property(EndpointProperty.HostAndPort)};Username={UsernameReference};Password={PasswordParameter}");
+}
--- a/samples/maildev-mailkit/MailDev.Hosting/MailDevResource.cs
+++ b/samples/maildev-mailkit/MailDev.Hosting/MailDevResource.cs
@@ -1,6 +1,9 @@
 namespace Aspire.Hosting.ApplicationModel;
 
 /// <summary>Represents a MailDev container resource.</summary>
+/// <param name="name">The resource name.</param>
+/// <param name="username">The optional SMTP username parameter.</param>
+/// <param name="password">The SMTP password parameter.</param>
 [AspireExport]
 public sealed class MailDevResource(
     [ResourceName] string name,
@@ -32,5 +35,15 @@ public sealed class MailDevResource(
     /// <inheritdoc />
     public ReferenceExpression ConnectionStringExpression =>
         ReferenceExpression.Create(
-            $"Endpoint=smtp://{SmtpEndpoint.Property(EndpointProperty.HostAndPort)};Username={UsernameReference};Password={PasswordParameter}");
+            $"Endpoint=smtp://{SmtpEndpoint.Property(EndpointProperty.HostAndPort)};Username={UsernameReference};******");
+
+    /// <inheritdoc />
+    public IEnumerable<KeyValuePair<string, ReferenceExpression>> GetConnectionProperties() =>
+    [
+        new("Host", ReferenceExpression.Create($"{SmtpEndpoint.Property(EndpointProperty.Host)}")),
+        new("Port", ReferenceExpression.Create($"{SmtpEndpoint.Property(EndpointProperty.Port)}")),
+        new("Username", UsernameReference),
+        new("Password", ReferenceExpression.Create($"{PasswordParameter}")),
+        new("Uri", ReferenceExpression.Create($"smtp://{SmtpEndpoint.Property(EndpointProperty.HostAndPort)}")),
+    ];
 }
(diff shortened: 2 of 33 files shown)


samples/maildev-mailkit/MailDev.Hosting/MailDevResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+        IResourceBuilder<ParameterResource>? password = null)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+
+        var passwordParameter = password?.Resource ??
+            ParameterResourceBuilderExtensions.CreateDefaultPasswordParameter(
+                builder, $"{name}-password");
+        var resource = new MailDevResource(name, username?.Resource, passwordParameter);
+
+        return builder.AddResource(resource)
```
adamint, line 38: Since this sample is marked run-only, could we exclude MailDev from the manifest? I confirmed the pinned head publishes it as `container.v0` with its SMTP credential parameter, which teaches the wrong lifecycle for a development mail sink. `.ExcludeFromManifest()` plus a manifest assertion should cover it.

samples/maildev-mailkit/MailDev.Hosting/MailDevResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+            .WithImage(Image)
+            .WithImageRegistry(Registry)
+            .WithImageTag(Tag)
+            .WithHttpEndpoint(
+                targetPort: 1080,
+                port: httpPort,
+                name: MailDevResource.HttpEndpointName)
```
adamint, line 45: `WaitFor(maildev)` only waits for the container to reach Running here because the resource has no health check. MailDev exposes `/healthz`; could we add an HTTP health check so the newsletter service waits for actual readiness rather than racing SMTP startup?

samples/maildev-mailkit/NewsletterService/NewsletterService.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
```
adamint, line 1: Could we add a `.slnx` for this sample containing the workload, integrations, AppHost, and tests? Repo CI discovers solutions or AppHosts, and this TypeScript sample path only restores ATS + runs `npm run aspire:build`; `NewsletterService`, `ServiceDefaults`, `MailKit.Client`, the C# AppHost, and the test projects are otherwise not compiled/tested by CI.

samples/maildev-mailkit/MailKit.Client/MailKitExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+            builder.Services.AddOpenTelemetry()
+                .WithMetrics(metrics => metrics.AddMeter(MailKit.Telemetry.SmtpClient.MeterName));
+        }
+    }
+
+    private static MailKitClientFactory CreateFactory(MailKitClientSettings settings)
+    {
+        if (settings.Endpoint is null)
+        {
+            throw new InvalidOperationException("The MailKit SMTP endpoint is not configured.");
```
adamint, line 104: Could we avoid throwing while constructing the health check dependency? Running `NewsletterService` without `ConnectionStrings:maildev`, `GET /health` returns an unhandled 500 from this line before `MailKitHealthCheck.CheckHealthAsync` can convert it to Unhealthy. The existing endpoint validation in `GetSmtpClientAsync` lets the health check report correctly.

samples/maildev-mailkit/NewsletterService/Program.cs
```diff
@@ -0,0 +1,61 @@
...
+})
+.WithName("Unsubscribe");
+
+app.Run();
+
+static MimeMessage CreateMessage(string email, string subject, string body)
+{
+    var message = new MimeMessage();
+    message.From.Add(new MailboxAddress("Aspire Newsletter", "newsletter@example.com"));
+    message.To.Add(MailboxAddress.Parse(email));
```
adamint: Could we validate with `MailboxAddress.TryParse` and return 400? I sent `{"email":"bad@@example.com"}` against the pinned service and this throws `MimeKit.ParseException`, returning an empty 500; `{}` and an empty value do the same.

samples/maildev-mailkit/MailDev.Hosting/MailDevResource.cs
```diff
@@ -0,0 +1,36 @@
...
+            : ReferenceExpression.Create($"{DefaultUsername}");
+
+    /// <summary>Gets the MailDev SMTP endpoint.</summary>
+    public EndpointReference SmtpEndpoint =>
+        _smtpEndpoint ??= new(this, SmtpEndpointName);
+
+    /// <inheritdoc />
+    public ReferenceExpression ConnectionStringExpression =>
+        ReferenceExpression.Create(
+            $"Endpoint=smtp://{SmtpEndpoint.Property(EndpointProperty.HostAndPort)};Username={UsernameReference};Password={PasswordParameter}");
```
adamint: Since this is an integration-authoring sample, could we implement `GetConnectionProperties()` for Host, Port, Username, Password, and Uri? The current default is empty, so generated/polyglot consumers only see the opaque connection string and miss the structured pattern we are trying to teach.

samples/maildev-mailkit/MailDev.Hosting/MailDev.Hosting.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <EnableAspireIntegrationAnalyzers>true</EnableAspireIntegrationAnalyzers>
```
adamint, line 7: These XML comments do not reach the generated ATS SDK because this project does not emit `MailDev.Hosting.xml` (confirmed in the build output). Could we enable `GenerateDocumentationFile` and inspect or assert the generated `.d.ts` JSDoc?

adamint (approved):
Reviewed pinned head `7ff86d2d6fd5ba479889cfa5a789f641ba434596`. The overall shape is sound and these are fixable in place, so I am approving. I think the must-fix items are keeping MailDev run-only, adding real readiness, making repo CI compile/test the .NET projects, and returning Unhealthy instead of an unhandled 500 for missing MailKit configuration. I also left comments on request validation, connection properties, and ATS docs.

The PR head advanced while this review ran; this decision is intentionally attached to the requested pinned commit.

samples/maildev-mailkit/MailKit.Client/MailKitExtensions.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            builder.Services.AddKeyedScoped<MailKitClientFactory>(
+                serviceKey,
+                (_, _) => CreateFactory(settings));
+        }
+
+        if (!settings.DisableHealthChecks)
+        {
+            builder.Services.AddHealthChecks().Add(new HealthCheckRegistration(
+                serviceKey is null ? "MailKit" : $"MailKit_{connectionName}",
```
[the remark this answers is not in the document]
adamint: Thanks for fixing the duplicate names. I think this is only half fixed: both unkeyed checks still call `GetRequiredService<MailKitClientFactory>()`, so the last registration wins. A focused test resolves both `MailKit_primary` and `MailKit_secondary` to `smtp://localhost:1026/`. Could we bind each check to its connection and assert the endpoint here, rather than just the label?

adamint (commented):
Incremental review of `f42053e`: the MailKit naming fix is incomplete, so I reopened that thread with a failing focused test. Ubuntu is failing in the unrelated Angular sample (`typescript@7.0.2` vs `@angular-devkit/build-angular` requiring `<6.1`), which #1866 already fixed on main. This head is still conflicting with main and red, so I’m leaving a comment rather than approving. Please merge main, address the MailKit thread, and rerun CI.

adamint (approved):
Approving this as the remaining issue is fixable in place. The health-check thread is still must-fix before merge, and this branch needs updating from main to clear the unrelated Angular failure and workflow conflict.

samples/maildev-mailkit/MailDev.Hosting/MailDevResourceBuilderExtensions.cs
IEvangelist: Done — the `AddMailDev` chain now ends with `.ExcludeFromManifest()`, plus a test asserting the resource carries a single `ManifestPublishingCallbackAnnotation` with a null callback.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

samples/maildev-mailkit/NewsletterService/NewsletterService.csproj
IEvangelist: Added `MailDevMailKit.slnx` containing all seven projects — the integrations, the C# AppHost, `NewsletterService`, `ServiceDefaults`, `MailKit.Client`, and both test projects. CI discovers the `.slnx` and restores/builds it in the sample's own directory, so the projects that were previously never compiled by CI now are. `dotnet test` on it runs all 18 tests green locally.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

samples/maildev-mailkit/MailKit.Client/MailKitExtensions.cs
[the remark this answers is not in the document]
IEvangelist: You're right, that was only half fixed — both non-keyed checks were still resolving the same factory, so the last registration won. Each check now gets a `MailKitClientFactory` constructed from its own connection settings inside the registration, so `MailKit_primary` and `MailKit_secondary` bind to distinct endpoints. Added a test that resolves both and asserts `smtp://localhost:1025` vs `:1026` on each check's factory endpoint.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

samples/maildev-mailkit/MailDev.Hosting/MailDevResourceBuilderExtensions.cs
IEvangelist: Added `.WithHttpHealthCheck("/healthz", endpointName: MailDevResource.HttpEndpointName)` to `AddMailDev`, so anything that `waitFor(maildev)` now blocks on real readiness instead of just the container reaching Running. Added a test asserting the `HealthCheckAnnotation` is present.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

samples/maildev-mailkit/NewsletterService/Program.cs
IEvangelist: Done — both `/subscribe` and `/unsubscribe` now validate the recipient with `MailboxAddress.TryParse` (guarding null/empty first) and return `400 Bad Request` for a missing or malformed address instead of throwing `MimeKit.ParseException` and returning an empty 500.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

samples/maildev-mailkit/MailDev.Hosting/MailDevResource.cs
IEvangelist: Implemented `GetConnectionProperties()` returning Host, Port, Username, Password, and Uri as structured `ReferenceExpression`s, so polyglot/generated consumers get the structured shape rather than only the opaque connection string. Added a test asserting each value expression (e.g. `Host` → `{maildev.bindings.smtp.host}`, `Uri` → `smtp://{...host}:{...port}`).

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

samples/maildev-mailkit/MailDev.Hosting/MailDev.Hosting.csproj
IEvangelist: Enabled `GenerateDocumentationFile`, so `MailDev.Hosting.xml` is now emitted and picked up by the ATS codegen. I checked the generated `.aspire/modules/aspire.mts`: `addMailDev` now carries its summary plus `@param`/`@returns`, and `MailDevResource` shows `/** Represents a MailDev container resource. */`. I also documented the primary-constructor params so generation stays warning-clean under `TreatWarningsAsErrors`.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

samples/maildev-mailkit/MailKit.Client/MailKitExtensions.cs
IEvangelist: Fixed — the registration now constructs the factory directly instead of resolving it from DI, and the factory constructor doesn't throw, so `GET /health` with no `ConnectionStrings:maildev` reports Unhealthy from inside `CheckHealthAsync` rather than an unhandled 500. Added a test asserting Unhealthy on the missing-endpoint path.

> [!NOTE]
> Auto-resolved by the [GitHub Copilot app](https://github.com/github/app).

